PR #15, opened 2020-12-14
updated barcode format to new specification

eli88fine:


tests/server/test_server_calling_routes.py
```diff
@@ -486,6 +501,58 @@ def test_start_recording__returns_error_code_and_message_if_barcode_is_invalid(
...
+            "ME200440001",
+            "allows header 'ME'",
+        ),
+        (
+            "MB200440001",
+            "allows header 'MB'",
+        ),
+    ],
+)
+def test_start_recording__allows_years_correct_barcode_headers(
```
tannermpeterson, line 545: Looks like this test name accidentally got messed up. I think it was meant to be: `test_start_recording__allows_correct_barcode_headers`

tannermpeterson (approved):

tannermpeterson:
Also, was the coverage <100% before this PR?

tests/server/test_server_calling_routes.py
eli88fine: fixed.  yes, there's a few places where "else is never true" and things like that in the refactor that I am putting off until the beta1.7 release. there aren't any actual lines that aren't covered.

PR #29, opened 2020-12-22
Stop/restart recording bug fix

tannermpeterson:


tests/test_file_writer.py
```diff
@@ -486,62 +486,109 @@ def test_FileWriterProcess__start_recording__sets_stop_recording_timestamp_to_no
...
-        "well_index": expected_well_idx,
+    data_packet = {
         "is_reference_sensor": False,
-        "data": np.array([[0], [1]], dtype=np.int32),
+        "well_index": expected_well_idx,
+        "data": np.array([[440000], [0]], dtype=np.int32),
     }
-    put_object_into_queue_and_raise_error_if_eventually_still_empty(
-        a_data_packet, board_queues[0][0], timeout_seconds=QUEUE_CHECK_TIMEOUT_SECONDS
+    board_queues[0][0].put(data_packet)
```
eli88fine, line 522: prefer `put_object_into_queue_and_raise_error_if_eventually_still_empty` imported from `.helpers` so that the `timeout_seconds` is already set to the test-suite-wide constant

tests/test_file_writer.py
```diff
@@ -486,62 +486,109 @@ def test_FileWriterProcess__start_recording__sets_stop_recording_timestamp_to_no
...
 
     assert stop_timestamps[0] is None
 
-    put_object_into_queue_and_raise_error_if_eventually_still_empty(
-        GENERIC_STOP_RECORDING_COMMAND,
-        from_main_queue,
-        timeout_seconds=QUEUE_CHECK_TIMEOUT_SECONDS,
+    this_command = copy.deepcopy(GENERIC_STOP_RECORDING_COMMAND)
+    this_command["timepoint_to_stop_recording_at"] = 2968000
+    from_main_queue.put(this_command)
```
eli88fine, line 537: same comment as above

tests/test_file_writer.py
```diff
@@ -486,62 +486,109 @@ def test_FileWriterProcess__start_recording__sets_stop_recording_timestamp_to_no
...
-    assert stop_timestamps[0] == 302412 * 125
+    assert stop_timestamps[0] == 2968000
 
-    to_main_queue.get(
-        timeout=QUEUE_CHECK_TIMEOUT_SECONDS
-    )  # pop off the initial receipt of start command message
-    assert_queue_is_eventually_not_empty(to_main_queue)
-    comm_to_main = to_main_queue.get(timeout=QUEUE_CHECK_TIMEOUT_SECONDS)
+    assert (
+        is_queue_eventually_of_size(
```
eli88fine, line 549: prefer `confirm_queue_is_eventually_of_size` for better failure messaging

tests/test_file_writer.py
```diff
@@ -486,62 +486,109 @@ def test_FileWriterProcess__start_recording__sets_stop_recording_timestamp_to_no
...
 
     assert file_writer_process.is_recording() is False
 
+    data_packet2 = {
+        "is_reference_sensor": False,
+        "well_index": expected_well_idx,
+        "data": np.array([[3760000], [0]], dtype=np.int32),
+    }
+    board_queues[0][0].put(data_packet2)
+    assert (
```
eli88fine, line 568: same as above about put+confirm

eli88fine (commented):
can you add an integration test for two full recording sessions and confirm both files contain data? it seems like that would be worthwhile to keep us safe in the future

tests/main/test_integration.py
```diff
@@ -584,6 +568,23 @@ def test_system_states_and_recorded_metadata_with_update_to_file_writer_director
...
                     )
 
+        # test second recording
+        for row_idx in range(4):
+            for col_idx in range(6):
+                with h5py.File(
+                    os.path.join(
+                        expected_recordings_dir,
+                        f"{expected_barcode2}__2020_06_15_141957",
+                        f"{expected_barcode2}__2020_06_15_141957__{chr(row_idx+65)}{col_idx+1}.h5",
```
eli88fine, line 578: not for future: use a function within `LabwareDefinition` to turn a row/column index into a well name.  It's unnecessarily risky to be re-making these calculations each time (risk of off-by-one errors and other things).  The root cause that Anna tracked down for the mislabeling of graphs in the SDK from image data was due to a bug in implementing one of these conversions instead of just using the `labware_domain_models` package

tests/test_file_writer.py
```diff
@@ -486,62 +487,86 @@ def test_FileWriterProcess__start_recording__sets_stop_recording_timestamp_to_no
...
 
-    to_main_queue.get(
-        timeout=QUEUE_CHECK_TIMEOUT_SECONDS
-    )  # pop off the initial receipt of start command message
-    assert_queue_is_eventually_not_empty(to_main_queue)
-    comm_to_main = to_main_queue.get(timeout=QUEUE_CHECK_TIMEOUT_SECONDS)
+    confirm_queue_is_eventually_of_size(
+        to_main_queue, 2, timeout_seconds=QUEUE_CHECK_TIMEOUT_SECONDS
+    )
+    to_main_queue.get_nowait()  # pop off the initial receipt of start command message
```
eli88fine, line 540: for future, we're stopping using `get_nowait` as it can cause sporadic test failures in CI.  prefer `get(timeout=QUEUE_CHECK_TIMEOUT_SECONDS)`

eli88fine (approved):

PR #32, opened 2020-12-28
Flask error mode

tannermpeterson:
Also cleaned up tests using functions from labware-domain-models


src/mantarray_desktop_app/server.py
```diff
@@ -202,8 +202,14 @@ def system_status() -> Response:
 
     Can be invoked by: curl http://localhost:4567/system_status
     """
-    board_idx = 0
     shared_values_dict = _get_values_from_process_monitor()
+    if (
+        "expected_software_version" in shared_values_dict
+        and shared_values_dict["expected_software_version"] != CURRENT_SOFTWARE_VERSION
+    ):
+        return Response(status="500 Versions of Electron and Flask EXEs do not match")
```
eli88fine, line 210: note for future - `500` is a catch-all code that Flask itself sometimes returns (i.e. if there was a python error while running the /system_status route code), so please update to something more specific within the 500 series

tests/main/test_integration.py
```diff
@@ -439,6 +440,8 @@ def test_system_states_and_recorded_metadata_with_update_to_file_writer_director
         # Tanner (6/15/20): Processes must be joined to avoid h5 errors with reading files, so clearing queues manually here in order to join
         test_process_manager.hard_stop_and_join_processes()
 
+        well_def = LabwareDefinition(row_count=4, column_count=6)
```
eli88fine, line 443: note for future - this can be declared at the top of the file as a module-level constant (as presumably it will be useful in other tests as well)---or maybe even as a constant in `fixtures.py` that could be imported across multiple tests.  maybe even within the desktop app repo `constants.py` itself. maybe even in the future it could be worthwhile having a few constants declared in the `labware-domain-models` package itself for immutable versions of very common formats like 24 well and 96 well and 384 well plates

eli88fine (approved):
looks good. some minor notes for future.

PR #35, opened 2020-12-30
Better integration tests

tannermpeterson:


tests/main/test_integration.py
```diff
@@ -123,22 +124,26 @@ def test_send_xem_scripts_command__gets_processed_in_fully_running_app(
     ]
     monitor_error_queue = test_process_monitor.get_fatal_error_reporter()
 
+    # Tanner (12/29/20): init with no bit file since Cloud9 does not have any real bit files and we are using a simulator
```
eli88fine, line 127: minor note: "repository" or "local dev environment" may be better than saying Cloud9 specifically.  Cloud9 is just a specific IDE that we happen to like using at the moment--we might change to another IDE in the future

tests/main/test_integration.py
```diff
@@ -123,22 +124,26 @@ def test_send_xem_scripts_command__gets_processed_in_fully_running_app(
...
     response = requests.get(
         f"{get_api_endpoint()}insert_xem_command_into_queue/initialize_board"
     )
     assert response.status_code == 200
 
+    # Tanner (12/29/20): Test xem_scripts will run using start up script. When this script complete the system will be in calibration_needed state
```
eli88fine, line 133: minor typo `completes`

tests/main/test_integration.py
```diff
@@ -687,6 +738,7 @@ def test_full_datapath(
         decimal=4,
     )
 
+    # Tanner (12/29/20): Good to do this at the end of tests to make sure they don't cause problems with other integration tests
     test_process_manager.hard_stop_and_join_processes()
     assert is_queue_eventually_empty(da_error_queue) is True
```
eli88fine, line 743: doesn't the error queue get drained as part of the `hard_stop` itself?  so asserting that it is empty after the hard stop would be always true even if there were an error? would it be better to check the dictionary returned by hard_stop to confirm there's nothing in the spot where error_queue items would be put for data_analyzer?

tests/main/test_integration.py
```diff
@@ -745,24 +802,27 @@ def test_app_shutdown__in_worst_case_while_recording_is_running(
...
         assert response.status_code == 200
-
         assert system_state_eventually_equals(RECORDING_STATE, 5) is True
 
         time.sleep(3)  # Tanner (6/15/20): This allows data to be written to files
 
+        # Tanner (12/29/20): Shutdown now that data is being acquired and actively written to files, which means each subprocess is doing something
         response = requests.get(f"{get_api_endpoint()}shutdown")
         assert response.status_code == 200
         # TODO (Eli 12/9/20): have /shutdown wait to return until the other processes have been stopped, or have a separate route to shut down other processes (that also waits)
```
eli88fine, line 812: ignore this `TODO` for now.  that's going to involve some more complicated interaction with the front-end that I don't want to dive into just yet

eli88fine (approved):

tests/main/test_integration.py
tannermpeterson: Yes this would be better

PR #44, opened 2021-01-07
Attack todos 2

tannermpeterson:


tests/test_data_analyzer.py
```diff
@@ -491,82 +508,32 @@ def test_DataAnalyzerProcess__drain_all_queues__drains_all_queues_except_error_q
...
-    )
-    assert (
-        is_queue_eventually_empty(
-            board_queues[0][0], timeout_seconds=QUEUE_CHECK_TIMEOUT_SECONDS
-        )
-        is True
-    )
+    confirm_queue_is_eventually_empty(from_main_queue)
+    confirm_queue_is_eventually_empty(to_main_queue)
+    confirm_queue_is_eventually_empty(board_queues[3][0])
```
eli88fine, line 532: 40 lines down to 6! Pretty!

tests/test_exe.py
```diff
@@ -76,7 +72,7 @@ def test_exe_can_access_xem_script_and_firmware_folders():
     response = requests.get(f"{get_api_endpoint()}shutdown")
     assert response.status_code == 200
 
-    time.sleep(5)  # wait for everything to fully shut down
+    time.sleep(7)  # wait for everything to fully shut down
```
eli88fine, line 75: definite need growing in the future for a refactor to how shutdown works to have it wait to return the response until shutdown is fully complete.  then it can either return 200 if everything soft stopped correctly, or some error code if it needed to hard stop things. then no need to sleep like this

eli88fine (commented):
looks good. once issue is resolved causing windows CI failures then I think we're good to merge

src/mantarray_desktop_app/process_monitor.py
```diff
@@ -88,8 +88,14 @@ def _check_and_handle_file_writer_to_main_queue(self) -> None:
         except queue.Empty:
             return
 
+        if "file_path" in communication:
+            communication["file_path"] = redact_sensitive_info_from_path(
+                communication["file_path"]
+            )
+        msg = f"Communication from the File Writer: {communication}".replace(
```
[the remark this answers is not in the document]
tannermpeterson: I think the `\\`s are created for some reason when the dict is converted to a string. I'm not sure of a better way to do this at the moment so I'll add a comment for now

tests/test_exe.py
tannermpeterson: I think we have a TODO for this

eli88fine (approved):

PR #56, opened 2021-01-14
Additional redaction

tannermpeterson:


src/mantarray_desktop_app/main.py
```diff
@@ -149,30 +149,32 @@ def main(
...
+    scrubbed_path_to_log_folder = redact_sensitive_info_from_path(path_to_log_folder)
 
     msg = f"Mantarray Controller v{CURRENT_SOFTWARE_VERSION} started"
     logger.info(msg)
     msg = f"Build timestamp/version: {COMPILED_EXE_BUILD_TIMESTAMP}"
     logger.info(msg)
     parsed_args_dict = copy.deepcopy(vars(parsed_args))
     del parsed_args_dict["initial_base64_settings"]
+    # Tanner (1/14/21): parsed_args_dict is only used to log the command line args at the moment, so log_file_dir can just be replaced here without affecting anything that actually needs the original value
+    if path_to_log_folder is not None:
```
eli88fine, line 161: is this `if` statement actually needed?  `redact_sensitive_info_from_path` already returns None if the input is None, so not having the if statement seems like it would still be fine.   In general, the less conditional logic in the code the easier it is to read

eli88fine (dismissed):
looks good. one minor note

src/mantarray_desktop_app/main.py
tannermpeterson: I'll quickly make that change

PR #61, opened 2021-01-21
More redactions of sensitive info from log files

tannermpeterson:


src/mantarray_desktop_app/process_monitor.py
```diff
@@ -112,8 +113,15 @@ def _check_and_handle_server_to_main_queue(self) -> None:
         except queue.Empty:
             return
 
+        msg: str
+        if "mantarray_nickname" in communication:
+            # Tanner (1/20/21): items in communication dict are used after this log message is generated, so need to creat a copy of the dict when redacting info
+            comm_copy = copy.deepcopy(communication)
+            comm_copy["mantarray_nickname"] = "*" * len(comm_copy["mantarray_nickname"])
```
eli88fine, line 120: note for future--it'd probably be good to refactor out a function something like `create_redacted_str(unredacted:str)->str` and also use that in the file path redaction and the tests.  That way if for some reason we ever want to change the character to something other than an asterisk...or if for some reason we think it is too sensitive to use the actual length of the original string, and we just want it to always put 10 asterisks, there would just be one place to change

src/mantarray_desktop_app/request_handler.py
```diff
@@ -0,0 +1,39 @@
...
+            scrubbed_msg += split_path[3] + split_path[4]
+
+            # Tanner (1/20/21): Tuples are immutable so need to convert to a list before modifying values, then convert back so they can be correctly passed to `_log`. Not sure if there's a better way to do this.
+            args_list = list(args)
+            args_list[0] = scrubbed_msg
+            args = tuple(args_list)
+
+        _log(
+            type_,
+            f"{self.address_string()} - - [{self.log_date_time_string()}] {message}\n",  # type: ignore  # Tanner (1/21/20): mypy is complaining that `address_string` is untyped
```
eli88fine, line 37: note for future: we already have a timestamp at the beginning of all our log messages.  so now that we have deeper control over the formatting, we could probably just not include this extra `log_date_time_string`

src/mantarray_desktop_app/server.py
```diff
@@ -972,10 +973,26 @@ def after_request(response: Response) -> Response:
         response = Response(status="404 Route not implemented")
     elif "get_available_data" in rule.rule and response.status_code == 200:
         del response_json["waveform_data"]["basic_data"]
+    elif "system_status" in rule.rule:
+        mantarray_nicknames = response_json.get("mantarray_nickname", {})
+        for board in mantarray_nicknames:
+            mantarray_nicknames[board] = "*" * len(mantarray_nicknames[board])
+    elif "set_mantarray_nickname" in rule.rule:
+        response_json["mantarray_nickname"] = "*" * len(
```
eli88fine, line 981: same not as above about a function to convert a sensitive item to the way we display it when redacted

eli88fine (commented):
looks good, but all the CI checks failed, so not sure what will need to be fixed to address that

eli88fine (approved):

setup.py
```diff
@@ -30,14 +30,13 @@
 
 setup(
     name="mantarray_desktop_app",
-    version="0.4.0",
+    # Eli (1/25/21): the version is now obtained from package.json during CI as the single source of truth
```
tannermpeterson, line 33: Does `pip install -e .` still work with the version removed?

tests/process_monitor/test_process_monitor_handling_communication_from_server.py
```diff
@@ -42,8 +42,6 @@ def test_MantarrayProcessesMonitor__check_and_handle_server_to_main_queue__handl
 ):
     monitor_thread, _, _, _ = test_monitor
 
-    test_process_manager.create_processes()
```
tannermpeterson, line 45: What made this line unnecessary in so many tests? Was it not necessary to begin with?

PR #65, opened 2021-01-26
Fixing more tests failing sporadically in CI

tannermpeterson:


tests/test_file_writer.py
```diff
@@ -216,15 +216,11 @@ def test_FileWriterProcess_soft_stop_not_allowed_if_command_from_main_still_in_q
...
     this_command["active_well_indices"] = [1]
     from_main_queue.put(this_command)
-    from_main_queue.put(this_command)
-    assert (
-        is_queue_eventually_of_size(
-            from_main_queue, 2, timeout_seconds=QUEUE_CHECK_TIMEOUT_SECONDS
-        )
-        is True
-    )
+    from_main_queue.put(copy.deepcopy(this_command))
```
eli88fine, line 219: note for future--it might be good to add a comment about why you think the 2nd time you put it in it should be deepcopied

eli88fine (approved):

PR #72, opened 2021-01-27
added barcode trimming to invalid barcodes sent to process_monitor

tannermpeterson:


src/mantarray_desktop_app/process_monitor.py
```diff
@@ -38,6 +38,7 @@
 from .exceptions import UnrecognizedRecordingCommandError
 from .process_manager import MantarrayProcessesManager
 from .server import ServerThread
+from .utils import _trim_barcode
```
eli88fine, line 41: minor note fore future (i'm actually surprised pylint doesn't complain about this)...I thought that you should never be importing something with a leading underscore. if it's being imported, that means it isn't 'private'

eli88fine (approved):
looks good--minor note

eli88fine (approved):

PR #85, opened 2021-01-29
Removing wait times within process for polling queues

eli88fine:
Having these wait times seems detrimental to the overall ability to measure performance of the subprocesses...everything is reporting 100% because the polling time is longer than the generic "sleep before iterating again" time.  I think with the updates we've made to the multiprocessing testing framework, we should be able to make the tests themselves be more robust to avoid needing to have these internal wait times.

I got things out of FileWriter, but wanted to get this merged before attempting the others.

### DIFF
--- a/tests/test_file_writer.py
+++ b/tests/test_file_writer.py
@@ -1535,6 +1535,9 @@ def test_FileWriterProcess__ignores_commands_from_main_while_finalizing_files_af
         }
         board_queues[0][0].put(final_ref_data_packet)
     confirm_queue_is_eventually_of_size(board_queues[0][0], 30)
+    time.sleep(
+        QUEUE_CHECK_TIMEOUT_SECONDS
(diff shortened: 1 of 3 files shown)


tests/test_file_writer.py
```diff
@@ -1195,11 +1190,13 @@ def test_FileWriterProcess__deletes_recorded_well_data_after_stop_time(
...
         ok_board_queues[0][0].put(data_packet)
-    assert is_queue_eventually_of_size(
+    confirm_queue_is_eventually_of_size(
         ok_board_queues[0][0],
         expected_remaining_packets_recorded + dummy_packets,
-        timeout_seconds=QUEUE_CHECK_TIMEOUT_SECONDS,
     )
+    # time.sleep(
+    #     QUEUE_CHECK_TIMEOUT_SECONDS
+    # )  # Eli (2/1/21): Even though the queue size has been confirmed in the above line, this extra sleep appears necessary to ensure that the subprocess can pull from the queue consistently using `get_nowait`. Not sure why this is required.
```
tannermpeterson, line 1199: Commented code

tannermpeterson (approved):

PR #88, opened 2021-02-03
Added base functionality of Mantarray MCU Simulator

tannermpeterson:


tests/test_mc_simulator.py
```diff
@@ -0,0 +1,303 @@
...
+    expected_1 = test_item[:-test_size_diff_1]
+    actual_1 = simulator.read(len(test_item) - test_size_diff_1)
+    assert actual_1 == expected_1
+
+    expected_2 = test_item[-test_size_diff_1:-test_size_diff_2]
+    actual_2 = simulator.read(test_size_diff_1 - test_size_diff_2)
+    assert actual_2 == expected_2
+
+    expected_3 = test_item[-1:]
+    actual_3 = simulator.read(1)
```
eli88fine, line 269: is this `1` coming from `test_size_diff_2`? or something else?  just trying to avoid magic numbers

tests/test_mc_simulator.py
```diff
@@ -0,0 +1,303 @@
...
+
+@pytest.mark.slow
+def test_MantarrayMCSimulator__handles_reads_of_size_less_than_next_packet_in_queue__when_simulator_is_running(
+    mantarray_mc_simulator,
+):
+    _, output_queue, _, testing_queue, simulator = mantarray_mc_simulator
+
+    test_item_1 = b"12345"
+    test_item_2 = b"67890"
+    testing_queue.put({"command": "add_read_bytes", "read_bytes": test_item_1})
```
eli88fine, line 281: see comment above about a `put` without a timeout

tests/test_mc_simulator.py
```diff
@@ -0,0 +1,303 @@
...
+    testing_queue.put({"command": "add_read_bytes", "read_bytes": test_item_2})
+    confirm_queue_is_eventually_of_size(testing_queue, 2)
+
+    # Tanner (2/2/20): this try/finally block ensures that the simulator is stopped even if the test fails. Problems can arise from processes not being stopped after tests complete
+    try:
+        simulator.start()
+        confirm_queue_is_eventually_empty(  # Tanner (2/2/21): Adding a large timeout here to avoid sporadic failures in CI (windows did not like a 5 second timeout)
```
eli88fine, line 288: if this is causing problems and it's not critical to this test, you could `invoke_process_run` twice prior to starting the simulator (assuming the goal of this test is just focused on the `.read` function and not the ability of the simulator to be populated with new data while runnning)

src/mantarray_desktop_app/mc_simulator.py
```diff
@@ -0,0 +1,120 @@
...
+                return bytes(0)
+            empty = True
+
+        if self._leftover_read_bytes is not None:
+            if empty:
+                next_packet = self._leftover_read_bytes
+            else:
+                next_packet = self._leftover_read_bytes + next_packet
+            self._leftover_read_bytes = None
+        if len(next_packet) > size:
```
eli88fine, line 113: what happens if `size > len(next_packet)`? I didn't see any unit tests for that possibility, but maybe I missed them

tests/test_mc_simulator.py
```diff
@@ -0,0 +1,303 @@
...
+            testing_queue, timeout_seconds=10
```
eli88fine, line 289: I'm not immediately sure why this long timeout is required (and why both items are still in the queue after the timeout).  maybe waiting for the subprocess to boot up before starting to confirm the status of the queue could potentially help (`is_start_up_complete`)?

eli88fine (commented):
Great foundation.  Some items about the way we test things that are worth considering

tests/test_mc_simulator.py
```diff
@@ -0,0 +1,420 @@
...
+    invoke_process_run_and_check_errors(simulator)
+    actual_item = simulator.read(size=len(expected_bytes))
+    assert actual_item == expected_bytes
+
+
+def test_MantarrayMCSimulator_read__returns_empty_bytes_if_no_bytes_to_read(
+    mantarray_mc_simulator,
+):
+    _, _, _, _, simulator = mantarray_mc_simulator
+    actual_item = simulator.read(size=10)
```
eli88fine, line 157: minor note: generally, we want to test the edge case, so just over the border of what would trigger the result we expect. the simplest way to test if no bytes are there is just to ask to read 1 byte

tests/test_mc_simulator.py
```diff
@@ -0,0 +1,420 @@
...
+):
+    spied_randint = mocker.spy(random, "randint")
+
+    _, _, _, _, simulator = mantarray_mc_simulator
+    mocker.patch.object(simulator, "get_cms_since_init", autospec=True, return_value=0)
+
+    invoke_process_run_and_check_errors(simulator)
+    spied_randint.assert_called_once_with(0, 10)
+
+    timestamp_bytes = bytes(8)
```
eli88fine, line 185: to make this a bit more clear (for example if we're looking back months from now during a refactor), I think this relates to mocking in line 180 right?  so there could be a variable `cms_since_init` and then you can set the `return_value` to that and then also say `timestamp_bytes = cms_since_init.to_bytes(8,byte_order='little')`

src/mantarray_desktop_app/mc_simulator.py
```diff
@@ -0,0 +1,158 @@
...
+
+        status_beacon = SERIAL_COMM_MAGIC_WORD_BYTES
+        status_beacon += num_bytes_in_content
+        status_beacon += self._get_cms_timestamp_bytes()
+        status_beacon += module_id
+        status_beacon += packet_type
+        status_beacon += _get_checksum_bytes(status_beacon)
+        if truncate:
+            trunc_index = random.randint(  # nosec B311 # Tanner (2/4/21): Bandit blacklisted this psuedo-random generator for security/cryptographic reasons which do not apply to the desktop app.
+                0, 10
```
eli88fine, line 94: why did you choose `10`? right now it's a bit of a magic number just looking at the code. If we are looking at the code in the future, it'd be good to know why.  Is this one less than the total length of the beacon?

tests/test_mc_simulator.py
```diff
@@ -0,0 +1,420 @@
...
+    test_items = [
+        {"command": "add_read_bytes", "read_bytes": test_item_1},
+        {"command": "add_read_bytes", "read_bytes": test_item_2},
+    ]
+    handle_putting_multiple_objects_into_empty_queue(test_items, testing_queue)
+
+    # Tanner (2/2/20): this try/finally block ensures that the simulator is stopped even if the test fails. Problems can arise from processes not being stopped after tests complete
+    try:
+        simulator.start()
+        confirm_queue_is_eventually_empty(testing_queue, timeout_seconds=5)
```
eli88fine, line 351: is the 5 second timeout necessary here? or can we use the standard timeout. it could potentially be confusing in the future trying to understand why it's set to 5 seconds and how that might impact a refactor

tests/test_mc_simulator.py
```diff
@@ -0,0 +1,420 @@
...
+
+    expected_2 = item_2[test_size_diff:]
+    actual_2 = simulator.read(
+        size=len(expected_2) + 1
+    )  # reading one more byte than are present
+    assert actual_2 == expected_2
+
+    expected_3 = bytes(0)
+    actual_3 = simulator.read(
+        size=10
```
eli88fine, line 394: see comment above about testing just over the edge of what would trigger the desired result.  do you see a benefit of testing 10 instead of 1 byte read size?

tests/test_mc_simulator.py
```diff
@@ -0,0 +1,420 @@
...
+    test_dict = {"command": "add_read_bytes", "read_bytes": test_item}
+    put_object_into_queue_and_raise_error_if_eventually_still_empty(
+        test_dict, testing_queue
+    )
+    invoke_process_run_and_check_errors(simulator)
+
+    expected_1 = test_item
+    actual_1 = simulator.read(
+        size=len(test_item) + 3
+    )  # reading arbitrarily greater number of bytes than are present
```
eli88fine, line 414: see note above about `+3` vs `+1`

eli88fine (commented):
great improvements. a few more thoughts/questions I had.  just trying to make sure we've laid the best foundation possible as we're figuring out how we want to go about testing this new simulator as we add functionality

tests/fixtures_mc_simulator.py
```diff
@@ -0,0 +1,45 @@
...
+        output_queue,
+        error_queue,
+        testing_queue,
+        read_timeout_seconds=SECONDS_TO_WAIT_WHEN_POLLING_QUEUES,
+    )
+
+    yield input_queue, output_queue, error_queue, testing_queue, simulator
+
+
+class MantarrayMCSimulatorNoBeacons(MantarrayMCSimulator):
```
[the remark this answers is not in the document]
eli88fine: I wonder if it'd be better to have the overridden beacon function put `bytes(0)` into the queue instead of doing absolutely nothing.  I don't think that would affect the results of reading from the queue, but it could expose unintended consequences of populating the queue.  Just trying to see how closely we can mimic the real code without overburdening the process of writing the tests

tests/fixtures_mc_simulator.py
```diff
@@ -17,15 +18,16 @@ def fixture_mantarray_mc_simulator():
...
     )
 
     yield input_queue, output_queue, error_queue, testing_queue, simulator
 
 
 class MantarrayMCSimulatorNoBeacons(MantarrayMCSimulator):
-    def _handle_status_beacon(self) -> None:
-        return
+    def _send_status_beacon(self, truncate=False) -> None:
+        self._time_of_last_status_beacon_secs = time.perf_counter()
```
eli88fine, line 29: thought for future---if `perf_counter` is imported from the `mc_simulator.py` module instead of `time`, then mocking it might have more of the expected effect.....maybe...I'm still not 100% sure exactly how namespaces work with mocking

eli88fine (approved):

PR #90, opened 2021-02-10
Comm module renaming

tannermpeterson:


src/mantarray_desktop_app/instrument_comm.py
```diff
@@ -0,0 +1,81 @@
...
+        for iter_board in board_connections:
+            if iter_board is not None:
+                iter_board.hard_stop(timeout=timeout)
+        return return_value
+
+    def determine_how_many_boards_are_connected(self) -> int:
+        # pylint: disable=no-self-use # currently a place holder just being mocked
+        return 1  # place holder for linting
+
+    @abc.abstractmethod
```
eli88fine, line 70: my understanding of the purpose of the decorator is that if you create a subclass and don't override this method, then it will raise an error during compile (even if you don't ever actually invoke it).  I'm not sure why coverage doesn't automatically ignore things decorated as abstract...maybe that's a setting we can tweak in our `.coveragerc` file

src/mantarray_desktop_app/ok_comm.py
```diff
@@ -512,14 +511,11 @@ def __init__(
...
     ):
-        # pylint-disable: duplicate-code # needed for the type definition of the board_queues
-        super().__init__(fatal_error_reporter, logging_level=logging_level)
-        self._board_queues = board_queues
-        self._board_connections: List[Union[None, okCFrontPanel]] = [None] * len(
-            self._board_queues
-        )
-        self._suppress_setup_communication_to_main = (
-            suppress_setup_communication_to_main
+        super().__init__(
```
eli88fine, line 514: possible idea for the future to avoid a bunch of duplicate code warnings for `init`: in the parent, have init call a method `_initialize_class_attributes()`.  Then in the subclasses, all the specific things like `_is_first_managed_read` could be declared as class attributes and then set to their initial value in that `_initialize_class_attributes()` method

eli88fine: if an instance assigns a value to a class attribute using `self._my_attribute`, then (my understanding) it becomes an object specific to that instance, so we wouldn't need to worry about it crossing over during test cases (i.e. it's not actually acting like a singleton class attribute)

eli88fine (approved):
looks good. some minor notes for the future. also, at some point the type hints in processmanager will need to be updated to say the parent class instead of OkComm

PR #94, opened 2021-02-16
Simulator handshake response

tannermpeterson:


tests/test_mc_simulator.py
```diff
@@ -29,14 +37,34 @@
...
-    return data_bytes + crc32(data_bytes).to_bytes(4, byteorder="little")
+    return data_bytes + crc32(data_bytes).to_bytes(
+        CHECKSUM_LENGTH_BYTES, byteorder="little"
+    )
+
+
+def create_data_packet(
+    timestamp: int,
+    module_id: int,
+    packet_type: int,
```
eli88fine, line 56: should this just be a standalone (or static) function in `mc_simulator` (that gets called by `_send_data_packet`)?  it seems like there's a lot of duplicated code between this and the code in the simulator...so we're not really testing, we may just be copy/pasting code into the tests.  and then a specific single test that the function creates the expected bytes with just a single set of inputs?

tests/test_mc_simulator.py
```diff
@@ -419,3 +445,154 @@ def test_MantarrayMCSimulator__handles_reads_of_size_greater_than_next_packet_in
...
+    test_module_id = 254
+    test_handshake = create_data_packet(
+        dummy_timestamp,
+        test_module_id,
+        dummy_packet_type,
+        DEFAULT_SIMULATOR_STATUS_CODE,
+    )
+
+    simulator.write(test_handshake)
+    time.sleep(QUEUE_CHECK_TIMEOUT_SECONDS)  # let input queue get populated
```
eli88fine, line 470: did we want to have the simulator take a kwarg of how much time (default `None`) to sleep after a `.write`, and supply an argument during test cases when instantiating the simulator? or did we determine it was better to have it as `sleep` lines in the test cases...I don't recall the outcome of that discussion

tests/test_mc_simulator.py
```diff
@@ -419,3 +445,154 @@ def test_MantarrayMCSimulator__handles_reads_of_size_greater_than_next_packet_in
...
+
+    with pytest.raises(UnrecognizedSerialCommModuleIdError, match=str(test_module_id)):
+        invoke_process_run_and_check_errors(simulator)
+
+
+def test_MantarrayMCSimulator__raises_error_if_unrecognized_packet_type_sent_from_pc(
+    mantarray_mc_simulator_no_beacon, mocker
+):
+    mocker.patch(
```
eli88fine, line 479: do we have `mock_print` or `patch_print` as a fixture already somewhere? might be worthwhile

tests/test_mc_simulator.py
```diff
@@ -419,3 +445,154 @@ def test_MantarrayMCSimulator__handles_reads_of_size_greater_than_next_packet_in
...
+        test_packet_type,
+        DEFAULT_SIMULATOR_STATUS_CODE,
+    )
+
+    simulator.write(test_handshake)
+    time.sleep(QUEUE_CHECK_TIMEOUT_SECONDS)  # let input queue get populated
+
+    with pytest.raises(UnrecognizedSerialCommPacketTypeError) as exc_info:
+        invoke_process_run_and_check_errors(simulator)
+    assert str(SERIAL_COMM_MAIN_MODULE_ID) in str(exc_info.value)
```
eli88fine, line 499: `match` is a kwarg that the `pytest.raises` function accepts that you can put regular expressions or other things into to do this type of thing. there may be other ways it's more robust than this, or maybe just easier to read, not sure. but that's the pytest standard practice https://docs.pytest.org/en/stable/assert.html

eli88fine (approved):
Looks good---just some notes for future (or now if you want) to jot down

tests/test_mc_simulator.py
tannermpeterson: I usually use that, it just had problems detecting multiple int values for some reason

tests/test_mc_simulator.py
tannermpeterson: I'm not sure we ever had that discussion, but I think it could be nice to have the simulator sleep on it's own during test cases

PR #96, opened 2021-02-18
Added --no-load-firmware flag to allow flask server to run from cmd line in Cloud9

tannermpeterson:
Also improved simulator


src/mantarray_desktop_app/process_monitor.py
```diff
@@ -70,11 +70,13 @@ def __init__(
         ],  # pylint: disable=unsubscriptable-object
         the_lock: threading.Lock,
         boot_up_after_processes_start: bool = False,
+        skip_load_firmware_file: bool = False,
```
eli88fine, line 73: I forget where I read this, but somewhere I think I remember seeing that ideally kwarg bools should be true if it's not too much trouble. that generally leads to shorter variable names: `load_firmware_file=True`

src/mantarray_desktop_app/mc_simulator.py
```diff
@@ -57,6 +83,7 @@ class MantarrayMCSimulator(InfiniteProcess):
         fatal_error_reporter: a queue to report fatal errors back to the main process
         testing_queue: queue used to send commands to the simulator. Should only be used in unit tests
         read_timeout_seconds: number of seconds to wait until read is of desired size before returning how ever many bytes have been read. Timeout should be set to 0 unless a non-zero value is necessary for unit testing
+        sleep_after_write_seconds: number of seconds to sleep after writing data to the simulator. This should always be None except when necessary for unit testing
```
eli88fine, line 86: we don't ever want to have this sleep set to something other than `None` if we actually start the simulator right? this is just for simple unit tests when the simulator is not running, correct?

might be good in the future to override `start` and raise an error if this is set to something other than `None` (then call the `super`), so that we don't accidentally get into situations in our test cases where we have a fixture with this sleep time set, and then start it running for a more "integration"-level test.  When the simulator is running, I think we want to avoid the sleep so that it behaves as close as possible to the actual running simulator---I don't think we would need the sleep when it's running anyway

eli88fine (approved):
looks good---minor notes for the future.

also one other --- there's a TODO `# TODO (Eli 12/9/20): figure out why this won't run on entrypoint.py in Windows GitHub`. do you think that with your fix here that we should be able to take out that exe-only `mark` so that it runs on entrypoint.py as well?

tests/test_mc_simulator.py
```diff
@@ -190,10 +190,15 @@ def test_MantarrayMCSimulator_read__returns_empty_bytes_if_no_bytes_to_read(
     assert actual_item == expected_item
 
 
-def test_MantarrayMCSimulator_write__puts_object_into_input_queue(
-    mantarray_mc_simulator,
-):
-    input_queue, _, _, _, simulator = mantarray_mc_simulator
+def test_MantarrayMCSimulator_write__puts_object_into_input_queue__with_no_sleep_after_write():
```
eli88fine, line 193: if the title is `with_no_sleep_after_write`, should there be an assertion that sleep isn't being called?

eli88fine (approved):

src/mantarray_desktop_app/process_monitor.py
tannermpeterson: I may be misremembering but I thought we specifically wanted them to be False and that's why `boot_up_after_processes_start` in the line above is False. The cmd line arg for skipping boot up stores true when passed but we specifically flip that value on line 232 in `main.py` due to it defaulting to False here, even though the default behavior of the app is to automatically boot up after the processes start

eli88fine: it's a somewhat fuzzy guideline.  `boot-Up_after_processes_start` is definitely clearer and more concise than `skip_boot_up_after_processes_start`.  And I agree that the command line arg should be called `skip` since we need the default behavior to be "do run it".  

I was just specifically talking about this `skip_load_firmware_file` vs `load_firmware_file` in this method here.  But it's not a big deal.  Just something to consider

PR #97, opened 2021-02-19
Instrument comm renaming

tannermpeterson:


src/mantarray_desktop_app/process_manager.py
```diff
@@ -36,7 +36,7 @@ def __init__(
     ) -> None:
         self._queue_container: MantarrayQueueContainer
 
-        self._ok_communication_process: OkCommunicationProcess
+        self._instrument_communication_process: OkCommunicationProcess
```
eli88fine, line 39: Should this type definition be InstrumentProcess now? Or not ready to make that change yet

src/mantarray_desktop_app/process_monitor.py
```diff
@@ -444,14 +440,14 @@ def _commands_for_each_run_iteration(self) -> None:
...
-                process_manager.queue_container().get_communication_to_ok_comm_queue(0)
+            to_instrument_comm = process_manager.queue_container().get_communication_to_instrument_comm_queue(
+                0
             )
             barcode_poll_comm = {
                 "communication_type": "barcode_comm",
                 "command": "start_scan",
             }
-            to_ok_comm.put(barcode_poll_comm)
+            to_instrument_comm.put(barcode_poll_comm)
```
eli88fine, line 450: As part of switching the get to getnowait we should consider what to do with these puts that currently have no timeout

src/mantarray_desktop_app/queue_container.py
```diff
@@ -247,7 +247,7 @@ def get_data_analyzer_error_queue(
 
     # def _create_queues(self) -> None:
     #     """Create all the queues and assign to the instance variables."""
```
eli88fine, line 249: Can this whole commented section be deleted?

eli88fine (approved):
Looks good. Minor notes

src/mantarray_desktop_app/queue_container.py
tannermpeterson: I suppose it can, I didn't actually see that when I was replacing everything

src/mantarray_desktop_app/process_manager.py
tannermpeterson: I was planning to change that once we added the new comm process but I don't think there would be any problem doing that right now

PR #99, opened 2021-02-23
adding check

annamcraig:


eli88fine:
@annamcraig - the CI run failed.  Do you get automatic notifications that it failed since you created the Pull Request? I'm not exactly sure how the github notifications work

eli88fine (commented):
It looks like this has been added in the workflow before the test suite is actually run. Can you confirm?

eli88fine (commented):
Also, I think you'll need to put 'pipenv run' in front of the command for pre commit. See the other place in the workflow using pre-commit

eli88fine (dismissed):

eli88fine:
looks good, but I think it's out of date with `development` and may need a `rebase`

eli88fine (approved):

eli88fine:
interesting results of that run.  looks like some files/folders should be added to .gitignore (`src/firmware/*.bit`, `coverage.xml` `src/drivers/*.exe`

The `constants.py` is a bit trickier, since we do intentionally modify that, but it's in the "Insert Build Number" step.  I think right after `Run Jest Tests` may be a good place to put the `git-dirty` check

eli88fine (approved):

PR #122, opened 2021-03-04
Ok comm get nowait

tannermpeterson:


tests/main/test_launch.py
```diff
@@ -195,7 +195,6 @@ def test_main__logs_system_info__and_software_version_at_very_start(
 
 
 @pytest.mark.timeout(GENERIC_MAIN_LAUNCH_TIMEOUT_SECONDS)
-@pytest.mark.slow
```
eli88fine, line 198: why did you take the `slow` mark off of these two?

eli88fine (approved):
looks great. one question in the comments, but if that's intentional, then approved to merge

tests/main/test_launch.py
tannermpeterson: They were running fairly quickly in Cloud9. Not sure if they were accidentally marked slow in the first place or not, but I didn't think they needed the `slow` mark anymore

PR #123, opened 2021-03-04
Mc simulator reboot

tannermpeterson:


src/mantarray_desktop_app/mc_simulator.py
```diff
@@ -131,46 +145,80 @@ def _send_data_packet(
 
     def _commands_for_each_run_iteration(self) -> None:
         self._handle_test_comm()
+        # if _reboot_time_secs is not None, this means the simulator is in a "reboot" phase
+        if self._reboot_time_secs is not None:
+            secs_since_reboot = _get_secs_since_reboot_command(self._reboot_time_secs)
+            # if secs_since_reboot is less than the reboot duration, simulator is still in the 'reboot' phase. Commands from PC will be discared and status beacons will not be sent
```
eli88fine, line 151: would commands from PC actually be discarded?  or would they just wait in the USB wire until the Mantarray rebooted again?  Was this something you checked with Aster? I don't remember explicitly discussing this, but I may have forgotten

src/mantarray_desktop_app/mc_simulator.py
```diff
@@ -131,46 +145,80 @@ def _send_data_packet(
...
 
+    def _handle_reboot_completion(self) -> None:
+        self._reset_start_time()
+        self._reboot_time_secs = None
+        self._reset_status_code_bits()
+        self._send_status_beacon(truncate=False)
+
+    def _discard_comm_from_pc(self) -> None:
+        try:
+            self._input_queue.get_nowait()
```
eli88fine, line 167: if we do need to discard all commands from the PC during a reboot, would this accomplish that? this is only discarding a single command each cycle.  would it be a more accurate representation of the Mantarray to just drain the queue as part of `_handle_reboot_completion`?

src/mantarray_desktop_app/mc_simulator.py
```diff
@@ -131,46 +145,80 @@ def _send_data_packet(
...
-                    SERIAL_COMM_CHECKSUM_FAILURE_PACKET_TYPE,
-                    trimmed_comm_from_pc,
+                    SERIAL_COMM_SIMPLE_COMMAND_PACKET_TYPE,
                 )
             else:
-                raise UnrecognizedSerialCommPacketTypeError(
-                    f"Packet Type ID: {packet_type} is not defined for Module ID: {module_id}"
-                )
+                # TODO
+                raise NotImplementedError()
```
eli88fine, line 210: was it intentional to leave this `TODO` here?

src/mantarray_desktop_app/mc_simulator.py
```diff
@@ -131,46 +145,80 @@ def _send_data_packet(
 
     def _commands_for_each_run_iteration(self) -> None:
         self._handle_test_comm()
+        # if _reboot_time_secs is not None, this means the simulator is in a "reboot" phase
+        if self._reboot_time_secs is not None:
+            secs_since_reboot = _get_secs_since_reboot_command(self._reboot_time_secs)
+            # if secs_since_reboot is less than the reboot duration, simulator is still in the 'reboot' phase. Commands from PC will be discared and status beacons will not be sent
+            if secs_since_reboot < 5:
```
eli88fine, line 152: a note about this number `5` would be useful.  Right now, I remember that this comes from a conversation with Aster/Farshad about the maximum amount of time it would take to reboot a Mantarray, but if we look back at this code in a month we could easily forget why there's a `5` here.   i.e. whether it was something chosen due to queue multiprocessing timing issues, a semi-arbitrary number, or something specifically tied to the way the Mantarray hardware operates

tests/test_mc_simulator.py
```diff
@@ -610,3 +612,228 @@ def test_MantarrayMcSimulator__allows_status_bits_to_be_set_through_testing_queu
...
+    simulator.write(test_reboot_command)
+    invoke_process_run_and_check_errors(simulator)
+
+    # remove reboot response packet
+    invoke_process_run_and_check_errors(simulator)
+    expected_reboot_response = create_data_packet(
+        0,
+        SERIAL_COMM_MAIN_MODULE_ID,
+        SERIAL_COMM_SIMPLE_COMMAND_PACKET_TYPE,
+        bytes(0),
```
eli88fine, line 732: would the status here be expected to be `expected_status_code_bits` (since that's the status of the Mantarray when the reboot command is received by the Mantarray)? I thought that was the way the Mantarray worked, but maybe I'm misunderstanding how the test is working or how the Mantarray works

tests/test_mc_simulator.py
```diff
@@ -610,3 +612,228 @@ def test_MantarrayMcSimulator__allows_status_bits_to_be_set_through_testing_queu
...
+    spied_reset.assert_called_once()
+
+
+def test_MantarrayMcSimulator__reset_status_code_after_rebooting(
+    mantarray_mc_simulator_no_beacon, mocker
+):
+    simulator = mantarray_mc_simulator_no_beacon["simulator"]
+    testing_queue = mantarray_mc_simulator_no_beacon["testing_queue"]
+
+    reboot_duration = 5
```
eli88fine, line 696: should this `5` be tied to something in `constants.py` that's used in both the tests and the main code?  it looks like we have two sources of "truth" here instead of having a single source of truth

tests/test_mc_simulator.py
```diff
@@ -610,3 +612,228 @@ def test_MantarrayMcSimulator__allows_status_bits_to_be_set_through_testing_queu
...
+    assert handshake_response[-8:-4] == DEFAULT_SIMULATOR_STATUS_CODE
+
+
+def test_MantarrayMcSimulator__processes_testing_commands_during_reboot(
+    mantarray_mc_simulator_no_beacon, mocker
+):
+    simulator = mantarray_mc_simulator_no_beacon["simulator"]
+    testing_queue = mantarray_mc_simulator_no_beacon["testing_queue"]
+
+    reboot_duration = 5
```
eli88fine, line 756: see note about about `reboot_duration` and 'single source of truth'

tests/test_mc_simulator.py
```diff
@@ -610,3 +612,228 @@ def test_MantarrayMcSimulator__allows_status_bits_to_be_set_through_testing_queu
...
+        SERIAL_COMM_MAIN_MODULE_ID,
+        SERIAL_COMM_SIMPLE_COMMAND_PACKET_TYPE,
+        bytes([SERIAL_COMM_REBOOT_COMMAND_BYTE]),
+    )
+    simulator.write(test_reboot_command)
+    invoke_process_run_and_check_errors(simulator)
+
+    # remove reboot response packet
+    invoke_process_run_and_check_errors(simulator)
+    expected_reboot_response = create_data_packet(
```
eli88fine, line 777: note for future: this process of sending a reboot command and removing the response seems potentially common across tests and might be able to be extracted out as a reusable function

eli88fine (commented):
looks good.  some comments and clarifications/questions about the way the Mantarray itself works

src/mantarray_desktop_app/mc_simulator.py
tannermpeterson: Aster said she was unsure if the UART buffer is active during reboot but if it is we should assume for now that it will get flushed after rebooting and the command will never be seen

src/mantarray_desktop_app/mc_simulator.py
tannermpeterson: That was a mistake

tests/test_mc_simulator.py
tannermpeterson: This is the "command acknowledged"  message it returns before rebooting. After completing reboot it sends a status beacon with the status bits

eli88fine (approved):

PR #126, opened 2021-03-08
making Flask multi-threaded to prepare for long-polling

eli88fine:


src/mantarray_desktop_app/request_handler.py
```diff
@@ -31,9 +32,9 @@ def log(self, type_: str, message: str, *args: Any) -> None:
...
-            type_,
-            f"{self.address_string()} - - {message}\n",  # type: ignore  # Tanner (1/21/20): mypy is complaining that `address_string` is untyped
-            *args,
-        )
+        with threading.Lock():  # Since Flask is running multi-threaded mode, make sure to acquire a lock before logging so logs don't get garbled # Eli (11/3/20): still unable to test if lock was acquired.
+            _log(
+                type_,
+                f"{self.address_string()} - - {message}\n",  # type: ignore  # Tanner (1/21/20): mypy is complaining that `address_string` is untyped
+                *args,
+            )
```
tannermpeterson, line 40: Acquiring a lock that is only ever acquired by this thread won't prevent other log messages from being printed at the same time right? Shouldn't there be another place this lock is acquired in order to prevent logging issues?

PR #130, opened 2021-03-15
Spectron in ci

eli88fine:
Got Spectron working for Windows test in CI.

Needed to add a flag to the Flask boot-up to help do Linux debugging/testing of the full E2E exe in Cloud9


tests/main/test_launch.py
```diff
@@ -111,7 +111,7 @@ def test_main__redacts_log_file_dir_from_log_message_of_command_line_args(mocker
         )
 
         redacted_log_file_dir = redact_sensitive_info_from_path(expected_log_file_dir)
-        expected_msg = f"Command Line Args: {{'debug_test_post_build': True, 'log_level_debug': False, 'skip_mantarray_boot_up': False, 'port_number': None, 'log_file_dir': '{redacted_log_file_dir}', 'expected_software_version': None, 'no_load_firmware': False}}"  # Tanner (1/14/21): Double curly braces escape formatting in f-strings, although Cloud9's syntax highlighter does not seem to recognize this
+        expected_msg = f"Command Line Args: {{'debug_test_post_build': True, 'log_level_debug': False, 'skip_mantarray_boot_up': False, 'port_number': None, 'log_file_dir': '{redacted_log_file_dir}', 'expected_software_version': None, 'no_load_firmware': False, 'skip_software_version_verification': False}}"  # Tanner (1/14/21): Double curly braces escape formatting in f-strings, although Cloud9's syntax highlighter does not seem to recognize this
```
tannermpeterson, line 114: We should find a better way to test the command line args. I haven't thought too much about it, but every time we add a new arg we have to update this test and the one above.

tannermpeterson: And the test below too it looks like

tannermpeterson (approved):

PR #134, opened 2021-03-16
Mc comm parse data

tannermpeterson:


src/mantarray_desktop_app/constants.py
```diff
@@ -231,6 +231,8 @@
 SECONDS_TO_WAIT_WHEN_POLLING_QUEUES = 0.02  # Due to the unreliablity of the .empty() .qsize() methods in queues, switched to a .get(timeout=) approach for polling the queues in the subprocesses.  Eli (10/26/20): 0.01 seconds was still causing sporadic failures in Linux CI in Github, so bumped to 0.02 seconds.
 
 # Serial Communication Values
+SERIAL_COMM_BAUD_RATE = 4000000
```
eli88fine, line 234: note for future: typically slightly easier to read as `4e6` or whatever the exact syntax is--rather than counting a large number of zeros

src/mantarray_desktop_app/mc_comm.py
```diff
@@ -2,17 +2,67 @@
...
+def _get_formatted_utc_now() -> str:
+    return datetime.datetime.utcnow().strftime("%Y-%m-%d %H:%M:%S.%f")
+
+
+def _get_seconds_since_read_start(start: float) -> float:
+    return perf_counter() - start
+
+
+def _process_main_module_comm(comm_from_instrument: bytes) -> None:
+    packet_type = comm_from_instrument[SERIAL_COMM_PACKET_TYPE_INDEX]
```
eli88fine, line 49: is there already a check that if `comm_from_instrument` is shorter than this index that some error would be raised containing the whole packet so that we could easily debug the issue? or would we just get an `IndexError` here?

src/mantarray_desktop_app/mc_comm.py
```diff
@@ -2,17 +2,67 @@
...
+            + comm_from_instrument[
+                SERIAL_COMM_ADDITIONAL_BYTES_INDEX:-SERIAL_COMM_CHECKSUM_LENGTH_BYTES
+            ]
+        )
+        raise SerialCommIncorrectChecksumFromPCError(returned_packet)
+
+    if packet_type == SERIAL_COMM_STATUS_BEACON_PACKET_TYPE:
+        pass
+    else:
+        module_id = comm_from_instrument[SERIAL_COMM_MODULE_ID_INDEX]
```
eli88fine, line 62: similar question as above about whether just an `IndexError` would happen.  maybe a check early on that the packet is at least the minimum expected length where things are being accessed by constant indices within the bytearray?

src/mantarray_desktop_app/mc_comm.py
```diff
@@ -30,21 +80,117 @@ def __init__(self, *args: Any, **kwargs: Any):
             self._board_queues
         )
 
+    def _setup_before_loop(self) -> None:
+        msg = {
+            "communication_type": "log",
+            "message": f'Microcontroller Communication Process initiated at {datetime.datetime.utcnow().strftime("%Y-%m-%d %H:%M:%S.%f")}',
```
eli88fine, line 86: note for future: can you use `_get_formatted_utc_now` here?

src/mantarray_desktop_app/mc_comm.py
```diff
@@ -30,21 +80,117 @@ def __init__(self, *args: Any, **kwargs: Any):
...
+        }
+        to_main_queue = self._board_queues[0][1]
+        if not self._suppress_setup_communication_to_main:
+            to_main_queue.put(msg)
```
eli88fine, line 90: note for future: should only use `put` with a `timeout` kwarg or `put_nowait`, to avoid risk of hanging

src/mantarray_desktop_app/mc_comm.py
```diff
@@ -30,21 +80,117 @@ def __init__(self, *args: Any, **kwargs: Any):
...
+        self.create_connections_to_all_available_boards()
+
+        board_idx = 0
+        board = self._board_connections[board_idx]
+        if isinstance(
+            board, MantarrayMcSimulator
+        ):  # pragma: no cover  # Tanner (3/16/21): it's impossible to connect to any serial port in CI, so _setup_before_loop will always be called with a simulator and this if statement will always be true in pytest
+            # Tanner (3/16/21): Current assumption is that a live mantarray will be running by the time we connect to it, so starting simulator here and waiting for it to complete start upt
```
eli88fine, line 98: future note: type `upt`

src/mantarray_desktop_app/mc_comm.py
```diff
@@ -30,21 +80,117 @@ def __init__(self, *args: Any, **kwargs: Any):
...
+            board.start()
+            while not board.is_start_up_complete():
```
eli88fine, line 100: is there any concern about relentlessly pinging this?

src/mantarray_desktop_app/mc_comm.py
```diff
@@ -30,21 +80,117 @@ def __init__(self, *args: Any, **kwargs: Any):
...
+        to_main_queue = self._board_queues[0][1]
+        if not self._suppress_setup_communication_to_main:
+            to_main_queue.put(msg)
+        self.create_connections_to_all_available_boards()
+
+        board_idx = 0
+        board = self._board_connections[board_idx]
+        if isinstance(
+            board, MantarrayMcSimulator
+        ):  # pragma: no cover  # Tanner (3/16/21): it's impossible to connect to any serial port in CI, so _setup_before_loop will always be called with a simulator and this if statement will always be true in pytest
```
eli88fine, line 97: potential future consideration: subclass the pyserial interface so that it has a dummy `start` method and `is_start_up_complete` event (and the `start` sets the event)

maybe overkill, maybe helpful, not sure

src/mantarray_desktop_app/mc_comm.py
```diff
@@ -30,21 +80,117 @@ def __init__(self, *args: Any, **kwargs: Any):
...
+        to_main_queue = self._board_queues[0][1]
+        for i in range(num_boards_connected):
+            msg = {
+                "communication_type": "board_connection_status_change",
+                "board_index": i,
+            }
+
+            for name in list(list_ports.comports()):
+                name = str(name)
+                # Tanner (3/15/21): As long as the STM eval board is used in the Mantarray, it will show up as so and we can look for the Mantarray by checking for STM in the name
```
eli88fine, line 123: would it make sense to include `msg["message"] = name`? for logging/debugging?

src/mantarray_desktop_app/mc_comm.py
```diff
@@ -30,21 +80,117 @@ def __init__(self, *args: Any, **kwargs: Any):
...
+                    Queue(),
+                    Queue(),
+                    Queue(),
+                    Queue(),
+                )
+            self.set_board_connection(i, serial_obj)
+            # TODO Tanner (3/15/21): add serial number and nickname to msg
+            msg["is_connected"] = not isinstance(serial_obj, MantarrayMcSimulator)
+            msg["timestamp"] = _get_formatted_utc_now()
+            to_main_queue.put(msg)
```
eli88fine, line 147: for future: `put` being used without timeout

src/mantarray_desktop_app/mc_simulator.py
```diff
@@ -170,12 +154,8 @@ def _handle_comm_from_pc(self) -> None:
             return
 
         # validate checksum before handling the communication
-        expected_checksum = crc32(comm_from_pc[:-SERIAL_COMM_CHECKSUM_LENGTH_BYTES])
-        actual_checksum = int.from_bytes(
-            comm_from_pc[-SERIAL_COMM_CHECKSUM_LENGTH_BYTES:],
-            byteorder="little",
-        )
-        if actual_checksum != expected_checksum:
+        checksum_is_valid = validate_checksum(comm_from_pc)
```
eli88fine, line 157: not for future: consider the more pythonic EAFP pattern of raising (and then handling) an error in the rare event that the checksum doesn't match, rather than the Java-esque LBYL pattern in current use

src/mantarray_desktop_app/mc_simulator.py
```diff
@@ -111,6 +86,15 @@ def __init__(
         self._status_code_bits: bytes = bytes(0)
         self._reset_status_code_bits()
 
+    @property
+    def in_waiting(self) -> int:
+        if self._leftover_read_bytes is None:
+            try:
+                self._leftover_read_bytes = self._output_queue.get_nowait()
```
eli88fine, line 93: is there any chance of this behaving in unexpected ways in unit tests if the `read_timeout_seconds` is set to a non-zero value?

tests/fixtures_mc_comm.py
```diff
@@ -19,3 +21,22 @@ def fixture_four_board_mc_comm_process():
         "error_queue": error_queue,
     }
     yield items_dict
+
+
+@pytest.fixture(scope="function", name="patch_comports")
+def fixture_patch_comports(mocker):
+    comport = "COM1"
+    mocked_comports = mocker.patch.object(
+        mc_comm.list_ports, "comports", autospec=True, return_value=[f"STM ({comport})"]
```
eli88fine, line 30: would it be more robust to have this return a list of size larger than 1 (i.e. with maybe the first value not being a Mantarray but being some other COM device instead)? or do you think that adds minimal value to the robustness of the test suite?

tests/test_mc_comm.py
```diff
@@ -40,6 +61,54 @@ def test_McCommunicationProcess_super_is_called_during_init(mocker):
...
 
 
+@pytest.mark.slow
+@freeze_time("2021-03-16 13:05:55.654321")
+def test_McCommunicationProcess_setup_before_loop__connects_to_boards__and_sends_message_to_main(
+    four_board_mc_comm_process,
+):
+    mc_process = four_board_mc_comm_process["mc_process"]
+    board_queues = four_board_mc_comm_process["board_queues"]
+    assert mc_process.get_board_connections_list() == [None, None, None, None]
```
eli88fine, line 71: note for future: I think you can do `[None] * 4`

tests/test_mc_comm.py
```diff
@@ -40,6 +61,54 @@ def test_McCommunicationProcess_super_is_called_during_init(mocker):
...
+
+    assert_queue_is_eventually_not_empty(board_queues[0][1])
+    process_initiated_msg = board_queues[0][1].get(timeout=QUEUE_CHECK_TIMEOUT_SECONDS)
+    assert process_initiated_msg["communication_type"] == "log"
+    assert (
+        process_initiated_msg["message"]
+        == "Microcontroller Communication Process initiated at 2021-03-16 13:05:55.654321"
+    )
+    # simulator is automatically started by mc_comm during setup_before_loop, so need to stop it
+    populated_connections_list[0].hard_stop()
```
eli88fine, line 85: is `hard_stop` needed to take the extra time to attempt to drain all the queues? or would a regular `stop` be faster and still sufficient to not risk breaking pipes?

tests/test_mc_comm.py
```diff
@@ -90,6 +159,79 @@ def test_McCommunicationProcess_set_board_connection__sets_connect_to_mc_simulat
...
+    }
+
+    actual_message = board_queues[0][1].get(timeout=QUEUE_CHECK_TIMEOUT_SECONDS)
+    assert actual_message["communication_type"] == "board_connection_status_change"
+    assert actual_message["board_index"] == board_idx
+    assert actual_message["is_connected"] is True
+    assert actual_message["timestamp"] == "2021-03-15 13:05:10.121212"
+
+
+@freeze_time("2021-03-15 13:05:10.121212")
```
eli88fine, line 201: note for future: typically, we like to freeze the time at slightly different values in different tests, just to inject some variety and ensure that somehow a time value wasn't hardcoded in the main code

eli88fine (commented):
looks great so far.  review in progress. had to tap out for the night, but take a look at comments/questions to see if any are relevant to need changes or not

tests/test_mc_comm.py
```diff
@@ -40,6 +61,54 @@ def test_McCommunicationProcess_super_is_called_during_init(mocker):
...
+    populated_connections_list[0].hard_stop()
+
+
+@pytest.mark.slow
+def test_McCommunicationProcess_setup_before_loop__does_not_send_message_to_main_when_setup_comm_is_suppressed():
+    board_queues, error_queue = generate_board_and_error_queues(num_boards=4)
+    mc_process = McCommunicationProcess(
+        board_queues, error_queue, suppress_setup_communication_to_main=True
+    )
+    assert mc_process.get_board_connections_list() == [None, None, None, None]
```
eli88fine, line 94: for future: it seems like a list of 4 None's is coming up frequently. could maybe extract as a constant at the top of this test module `expected_empty_board_connections_list`

tests/test_mc_comm.py
```diff
@@ -40,6 +61,54 @@ def test_McCommunicationProcess_super_is_called_during_init(mocker):
...
+    # Other parts of the process after setup may or may not send messages to main, so drain queue and make sure none of the items (if present) have a setup message
+    to_main_queue_items = drain_queue(board_queues[0][1])
+    for item in to_main_queue_items:
+        if "message" in item:
+            assert (
+                "Microcontroller Communication Process initiated" not in item["message"]
+            )
+
+    # simulator is automatically started by mc_comm during setup_before_loop, so need to stop it
+    populated_connections_list[0].hard_stop()
```
eli88fine, line 109: similar question as above about whether ` stop` would be sufficient

tests/test_mc_comm.py
```diff
@@ -269,3 +434,201 @@ def test_McCommunicationProcess_register_magic_word__raises_error_if_search_exce
...
+    assert mc_process.is_registered_with_serial_comm(0) is False
+    assert mc_process.is_registered_with_serial_comm(1) is False
+    assert mc_process.is_registered_with_serial_comm(2) is False
+    assert mc_process.is_registered_with_serial_comm(3) is False
+
+
+def test_McCommunicationProcess__raises_error_if_magic_word_is_incorrect_in_packet_after_previous_magic_word_has_been_registered(
+    four_board_mc_comm_process, mantarray_mc_simulator_no_beacon, mocker
+):
+    mocker.patch(
```
eli88fine, line 454: note for future--could probably extract this as a common fixture `patched_print`

src/mantarray_desktop_app/mc_simulator.py
```diff
@@ -111,6 +86,15 @@ def __init__(
         self._status_code_bits: bytes = bytes(0)
         self._reset_status_code_bits()
 
+    @property
+    def in_waiting(self) -> int:
```
eli88fine, line 90: Is this always guaranteed to be accurate? What if `leftover_read_bytes` is `not None` but there are still things in the queue to be pulled out?

I'm not saying we need to guarantee its accuracy. But if it's not accurate, that should be clearly denoted in the docstring (with as much detail as possible about what it is guaranteed to do, and what situations would make it inaccurate).

The `in_waiting` for `pyserial` is guaranteed to be accurate right?

eli88fine (approved):
take a look at the comments/questions. If the answers to the questions are all that there's no changes needed, go ahead and merge

eli88fine (approved):

PR #140, opened 2021-03-19
MCU metadata

tannermpeterson:
This PR got out of hand due to other features that needed to be implemented before this specific feature. In future need to write smaller stories in addition to making sure all the functionality needed for the story is already implemented in McComm/McSimulator. Hopefully this will become less of an issue as these processes grow and have all of their base features implemented


src/mantarray_desktop_app/queue_utils.py
```diff
@@ -1,21 +0,0 @@
...
-from __future__ import annotations
-
-from queue import Queue
-from typing import Any
-from typing import List
-
-from stdlib_utils import safe_get
-
-
-def _drain_queue(
```
tannermpeterson, line 12: Removed this file since we can now move drain_queue into stdlib_utils

tests/mc_simulator/test_mc_simulator.py
```diff
@@ -846,3 +891,96 @@ def test_MantarrayMcSimulator__does_not_send_status_beacon_while_rebooting(
...
+    invoke_process_run_and_check_errors(simulator)
+
+    # Check that nickname is updated
+    actual_metadata = simulator.get_metadata_dict()
+    assert actual_metadata[MANTARRAY_NICKNAME_UUID.bytes] == convert_to_metadata_bytes(
+        expected_nickname
+    )
+    # Check that correct response is sent
+    expected_response = create_data_packet(
+        spied_get_cms_since_init.spy_return,
```
eli88fine, line 950: in the future, it may be good to figure out a way to test this that doesn't require checking the timestamp as well (unless you think that's critical to this unit of functionality)--as I imagine this may come up more frequently as new packet types are added

eli88fine (approved):
looks good. minor notes/questions

tests/mc_simulator/test_mc_simulator.py
tannermpeterson: I agree this is a good idea. In some cases we definitely want the timestamp, but in others we are mostly focused on testing the module ID, packet type, and main body of the data packet and possibly the checksum. If we think only one checksum related test is sufficient per module (one for packets sent from mc_comm and one for packets sent from the simulator) then we could definitely ignore the timestamp in all other tests

src/mantarray_desktop_app/mc_simulator.py
```diff
@@ -61,6 +73,22 @@ class MantarrayMcSimulator(InfiniteProcess):
         read_timeout_seconds: number of seconds to wait until read is of desired size before returning how ever many bytes have been read. Timeout should be set to 0 except in unit testing scenarios where necessary
     """
 
+    default_mantarray_nickname = "Mantarray Simulator (MCU)"
+    default_mantarray_serial_number = "M02001901"
+    default_pcb_serial_number = (
+        "TBD"  # TODO Tanner (3/17/21): implement this once the format is determined
+    )
+    default_firmware_version = "0.0.0"
+    default_metadata_values: Dict[UUID, Any] = {
```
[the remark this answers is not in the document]
tannermpeterson: The UUID/value version is useful for testing and the bytes/bytes version makes including metadata in data packets more efficient. I agree it should also be an immutabledict, and I left the type as Any since I didn't know how many values would be added. It can probably also just be changed to Union[str, int].

src/mantarray_desktop_app/mc_comm.py
```diff
@@ -141,14 +176,107 @@ def create_connections_to_all_available_boards(self) -> None:
...
+            else:
+                raise UnrecognizedCommandFromMainToMcCommError(
+                    f"Invalid command: {comm_from_main['command']} for communication_type: {communication_type}"
+                )
+        else:
+            raise UnrecognizedCommandFromMainToMcCommError(
+                f"Invalid communication_type: {communication_type}"
+            )
+        self._command_awaiting_response = comm_from_main
+        if not input_queue.empty():
```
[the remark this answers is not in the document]
tannermpeterson: If we're going to use a timeout, could we use something like `is_queue_eventually_empty`?

PR #142, opened 2021-03-23
MC Comm Send Handshake

tannermpeterson:
Also fixed the format of some other commands


src/mantarray_desktop_app/mc_comm.py
```diff
@@ -243,39 +258,33 @@ def _process_next_communication_from_main(self) -> None:
...
-            )  # Tanner (3/17/21): to be consistent with OkComm, command responses will be sent back to main after the command is acknowledged by the Mantarray
-            self._command_awaiting_response = None
-        else:
-            module_id = comm_from_instrument[SERIAL_COMM_MODULE_ID_INDEX]
-            raise UnrecognizedSerialCommPacketTypeError(
-                f"Packet Type ID: {packet_type} is not defined for Module ID: {module_id}"
-            )
+    def _handle_handshake(self) -> None:
+        board_idx = 0
+        board = self._board_connections[board_idx]
```
jperson, line 268: since `board` isn't used in this function except for the if statement could it be
```python
if not self._board_connections[board_idx]:
    return
```

also is there ever a chance that `self._board_connections` is `None`? If so this will raise.

src/mantarray_desktop_app/mc_comm.py
```diff
@@ -323,6 +332,39 @@ def _handle_incoming_data(self) -> None:
         else:
             raise UnrecognizedSerialCommModuleIdError(module_id)
 
+    def _process_main_module_comm(self, comm_from_instrument: bytes) -> None:
+        packet_body = comm_from_instrument[
+            SERIAL_COMM_ADDITIONAL_BYTES_INDEX:-SERIAL_COMM_CHECKSUM_LENGTH_BYTES
+        ]
+
+        packet_type = comm_from_instrument[SERIAL_COMM_PACKET_TYPE_INDEX]
```
jperson, line 340: is there any need for error checking on the length of `comm_from_instrument`?

src/mantarray_desktop_app/mc_comm.py
```diff
@@ -323,6 +332,39 @@ def _handle_incoming_data(self) -> None:
...
+        if packet_type == SERIAL_COMM_CHECKSUM_FAILURE_PACKET_TYPE:
+            returned_packet = SERIAL_COMM_MAGIC_WORD_BYTES + packet_body
+            raise SerialCommIncorrectChecksumFromPCError(returned_packet)
+
+        if packet_type == SERIAL_COMM_STATUS_BEACON_PACKET_TYPE:
+            pass  # TODO Tanner (3/17/21): Implement this in a story dedicated to parsing/handling errors codes in status beacons and handshakes
+        elif packet_type == SERIAL_COMM_COMMAND_RESPONSE_PACKET_TYPE:
+            response_data = packet_body[SERIAL_COMM_TIMESTAMP_LENGTH_BYTES:]
+            if len(self._commands_awaiting_response) == 0:
```
jperson, line 349: an empty collections structures are false values in python so you don't need to get the length
```python
if not self._commands_awaiting_respose:
    ...
```

src/mantarray_desktop_app/mc_comm.py
tannermpeterson: `self._board_connections` will always be a list, but it's items can be `None`

src/mantarray_desktop_app/mc_comm.py
tannermpeterson: There is, but it hasn't been determined yet what we should do if the length is less than the amount we tried to read. The way bytes are read from both the simulator and live board we will never get more bytes than we attempt to read. I put a TODO for this on line 307

jperson: 👍

src/mantarray_desktop_app/mc_comm.py
tannermpeterson: Also I'm not sure the exact cases we do this, but we've typically been using `if x is None:` instead of `if not x:` to prevent weird edge cases where the variable we're testing is falsey but not actually `None`. I don't think that necessarily applies to an instance of `MantarrayMcSimulator`, but I wrote the check that way to be consistent with how `ok_comm` checks for board connections

jperson: and it's always going to have at least 1 element?

jperson:
LGTM, just some nitpicks

jperson (approved):

src/mantarray_desktop_app/mc_comm.py
tannermpeterson: It will have exactly the number of maximum allowed board connections which is currently 4. We don't actually support connecting to 4 instruments at the moment (and possibly never will), this is just a placeholder for down the road when that may need to be supported

jperson: 👍
I just meant that `board` is only used to check if it's `None` or not so you could write it without the unused variable instead, it's up to you though.

src/mantarray_desktop_app/mc_comm.py
```diff
@@ -323,6 +332,39 @@ def _handle_incoming_data(self) -> None:
...
+                raise NotImplementedError(  # stop mypy complaining
+                    "_commands_awaiting_response shoudn't be empty here"
```
eli88fine, line 351: It's possible the firmware may go rogue and send out a response packet even when one wasn't requested (which could trigger this line, correct?).  It may be worth having a more detailed error here that would include the contents of the packet in the error message to help in debugging. Rather than just a simple `NotImplementedError`

tests/mc_comm/test_mc_comm.py
```diff
@@ -958,3 +927,62 @@ def test_McCommunicationProcess__processes_command_from_main_when_process_is_ful
             break
     else:
         assert False, "expected response to main not found"
+
+
+def test_MantarrayMcSimulator__sends_handshake_every_5_seconds__and_includes_correct_timestamp__and_processes_response(
```
eli88fine, line 932: is this a test of the Simulator, or McComm?

tests/mc_simulator/test_mc_simulator.py
```diff
@@ -768,9 +748,9 @@ def test_MantarrayMcSimulator__reset_status_code_after_rebooting(
     invoke_process_run_and_check_errors(simulator)
 
     # send reboot command
-    dummy_timestamp = 0
+    expected_timestamp = 11110
```
eli88fine, line 751: is this number arbitrary? related to something?

tests/mc_simulator/test_mc_simulator.py
```diff
@@ -816,9 +792,9 @@ def test_MantarrayMcSimulator__processes_testing_commands_during_reboot(
     )
 
     # send reboot command
-    dummy_timestamp = 0
+    expected_timestamp = 1000
```
eli88fine, line 795: it appears these numbers may just be arbitrary to inject some variety into the test suite. If they're truly arbitrary, a good way to indicate that (and inject even more variety) is to actually have it be a randomized number (within some range).  `pytest-randomly` lets us deterministically re-run a test suite with the same randomly generated numbers, since it controls the random seed, so that avoids any concern about not being able to debug a failing test

eli88fine (approved):
looks good. some minor notes/questions. unless you think any are serious, they can probably be addressed an quick fixes in the next PR

eli88fine (dismissed):

jperson (approved):

PR #145, opened 2021-03-29
Smoother release

eli88fine:
* got the VRT to ignore the minor and patch version numbers (still included the major version number to have some test that the font/color is correct and that the number is being created...and not some error code)
* Got the unit tests to dynamically extract the app version instead of having it hardcoded--so it doesn't have to be adjusted in the tests when the version changes
* Added a debugging param to the `/shutdown` call from the frontend when its called by the user closing the application

### DIFF
--- a/src/main/index.js
+++ b/src/main/index.js
@@ -154,7 +154,9 @@ app.on("will-quit", function () {
   console.log("will-quit event being handled"); // allow-log
   // mainWindow = null;
 
-  axios.get(`http://localhost:${flask_port}/shutdown`);
+  axios.get(
+    `http://localhost:${flask_port}/shutdown?called_through_app_will_quit=true`
+  );
 });
 
 // win_handler = require("./mainWindow");
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "MantarrayController",
-  "version": "0.4.2",
+  "version": "0.4.3",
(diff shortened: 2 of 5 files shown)


src/main/index.js
```diff
@@ -154,7 +154,9 @@ app.on("will-quit", function () {
   console.log("will-quit event being handled"); // allow-log
   // mainWindow = null;
 
-  axios.get(`http://localhost:${flask_port}/shutdown`);
+  axios.get(
+    `http://localhost:${flask_port}/shutdown?called_through_app_will_quit=true`
```
robinr, line 158: On the python flask server end the new query parameter 'called_through_app_will_quit', is it handled to shutdown ?

robinr (commented):
Flask server changes with query parameters, is it working as a normal shutdown ?

src/main/index.js
eli88fine: Flask allows you to pass extra unhandled parameters without having an effect. This will just help with logging, hopefully to help troubleshoot the issue one of our beta customers is having.  Similar to how when the frontend calls `/system_status`, it includes the `current_vuex_status_uuid` as a parameter---the backend doesn't use that, it's just for being able to debug better using the logs.

robinr:
No comments.

jperson (approved):

PR #147, opened 2021-03-30
Electron version detection fix

eli88fine:
`process.env.npm_package_version` should apparently NOT be used. It works in CI using the test runner, but does not actually work when running on a standalone machine--it just evaluates to undefined.


robinr (commented):
no comments

robinr (approved):

jperson (approved):

PR #155, opened 2021-04-05
Initial serial status codes

tannermpeterson:


src/mantarray_desktop_app/mc_comm.py
```diff
@@ -412,7 +425,14 @@ def _process_main_module_comm(self, comm_from_instrument: bytes) -> None:
                 )
             prev_command = self._commands_awaiting_response.popleft()
             if prev_command["command"] == "handshake":
-                # see note above: Tanner (3/17/21)
+                status_code = int.from_bytes(response_data, byteorder="little")
+                log_msg = f"Handshake response received from instrument. Status Code: {status_code}"
```
eli88fine, line 429: note for future---lines 408-419 seem substantively similar...could consider extracting out as a common function

src/mantarray_desktop_app/mc_simulator.py
```diff
@@ -71,6 +77,15 @@ def _get_secs_since_reboot_command(last_time: float) -> float:
...
 
+def _get_secs_since_boot_up(start_time: float) -> float:
+    return perf_counter() - start_time
+
+
+def _get_secs_since_last_comm_from_pc(last_time: float) -> float:
+    return perf_counter() - last_time
+
+
+# pylint: disable=too-many-instance-attributes
```
eli88fine, line 88: note for future: typically we provide some comment as a justification of why you think this situation warrants disabling one of the normal pylint rules

eli88fine: also, I think (not 100% sure) that if you put the comment up here, it disables it for the rest of the module.  If you want to disable it just for a class, I think you can put it right inside the class declaration (i.e. line 90)

tests/mc_simulator/test_mc_simulator.py
```diff
@@ -73,6 +79,10 @@
     bytes(0),
 )
 
+DEFAULT_SIMULATOR_STATUS_CODE = SERIAL_COMM_BOOT_UP_CODE.to_bytes(
+    SERIAL_COMM_STATUS_CODE_LENGTH_BYTES, byteorder="little"
```
eli88fine, line 83: this seems to come up frequently. for future could consider creating
`def convert_status_code_to_bytes(code:int)->bytes`

eli88fine (dismissed):
looks good, some minor notes for future

src/mantarray_desktop_app/mc_simulator.py
tannermpeterson: It looks like that's how we are disabling it in all of the other subprocesses

eli88fine (approved):

PR #156, opened 2021-04-06
Removed all instances of put in source and test files

tannermpeterson:


tannermpeterson:
Running 2 other instances of the Dev workflow on this branch to check for any sporadic failures that may come about from the changes since they pertain to queues

jperson:
lgtm, are all the queues set to have unlimited size?

tannermpeterson:
I'm not sure if they have unlimited size, but the docs seem to indicate that a max size can be specified which is something we don't do. There's nothing mentioned about a default size

eli88fine:
My understanding is that they have a really large size that we've never hit.  But we don't have any mechanism built in of what to do if for some reason we were filling up the buffer.  So we'd rather have it throw an error (by using `put_nowait`) so we immediately know that we have actually hit a limit and need to figure out a mechanism in the code to robustly handle the possibilty that the queue could actually get full

jperson:
> class queue.Queue(maxsize=0) Constructor for a FIFO queue. maxsize is an integer that sets the upperbound limit on the number of items that can be placed in the queue. Insertion will block once this size has been reached, until queue items are consumed. If maxsize is less than or equal to zero, the queue size is infinite.

The docs say they default to maxsize=0 which is unlimited, mostly asking because `put_nowait` will throw if the queue is full

jperson (approved):

eli88fine (approved):

PR #158, opened 2021-04-07
Serial comm set time

tannermpeterson:


src/mantarray_desktop_app/mc_simulator.py
```diff
@@ -183,8 +195,15 @@ def _send_data_packet(
         data_to_send: bytes = bytes(0),
         truncate: bool = False,
     ) -> None:
+        # TODO Tanner (4/7/21): convert timestamp to microseconds once real board makes the switch
+        timestamp = (
+            (self._baseline_time_usec + _perf_counter_us())
```
eli88fine, line 200: wouldn't `_perf_counter_us` need to be relative to the some timepoint? maybe I'm misunderstanding, but here wouldn't perf_counter just essentially be an arbitrary number?

eli88fine (approved):
some minor questions--if they're not relevant then looks good

src/mantarray_desktop_app/serial_comm_utils.py
```diff
@@ -143,3 +143,15 @@ def convert_to_status_code_bytes(status_code: int) -> bytes:
...
     )
+
+
+def convert_to_timestamp_bytes(timestamp: int) -> bytes:
+    return timestamp.to_bytes(SERIAL_COMM_TIMESTAMP_LENGTH_BYTES, byteorder="little")
+
+
+# Tanner (4/7/21): This method should not be used in the simulator. It has its own way of determining the timestamp to send in order to behave more accurately like the real Mantarray instrument
+def get_serial_comm_timestamp() -> int:
+    # TODO Tanner (4/7/21): change this to microseconds when the real Mantarray makes the switch
```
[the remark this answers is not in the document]
tannermpeterson: Looks like I forgot to convert to cms

src/mantarray_desktop_app/mc_simulator.py
tannermpeterson: Yes it does need to be relative to a start timepoint

tests/test_serial_comm_utils.py
```diff
@@ -158,7 +158,7 @@ def test_parse_metadata_bytes__returns_metadata_as_dictionary(
 @freeze_time("2021-04-07 13:14:07.234987")
 def test_get_serial_comm_timestamp__returns_microseconds_since_2021_01_01():
     expected_usecs = (
-        datetime.datetime.utcnow() - SERIAL_COMM_TIMESTAMP_EPOCH
+        datetime.datetime.now(tz=datetime.timezone.utc) - SERIAL_COMM_TIMESTAMP_EPOCH
```
eli88fine, line 161: did this not pass with `datetime.datetime.utcnow()`?

eli88fine (approved):
minor question

tests/test_serial_comm_utils.py
tannermpeterson: `utcnow` is a "naive" datetime object and cannot be compared to an "aware" datetime object https://docs.python.org/3/library/datetime.html#datetime.datetime.now

tannermpeterson: I believe this means aware of timezone offset

PR #161, opened 2021-04-09
Added dump EEPROM command

tannermpeterson:


Pipfile
```diff
@@ -33,7 +33,7 @@ immutable-data-validation = "==0.2.1"
 labware-domain-models = "==0.3.1"
 nptyping = "==1.4.0"
 numpy = "==1.20.1"
-mantarray-file-manager = "==0.4.5"
+mantarray-file-manager = "==0.4.8"
```
eli88fine, line 36: I think there may have been some new features and a new H5 file format introduced in `0.4.6`. I think there's an Asana Story already to incorporate these at some point, but just an extra note not to forget in case it isn't noted elsewhere

eli88fine (approved):

Pipfile
tannermpeterson: https://app.asana.com/0/1153227363135192/1199878181059874

PR #184, opened 2021-05-05
Beta 2 data streaming

tannermpeterson:
Also made fix in `convert_to_metadata_bytes`


src/mantarray_desktop_app/mc_simulator.py
```diff
@@ -204,14 +239,35 @@ def in_waiting(self) -> int:
                 return 0
         return len(self._leftover_read_bytes)
 
+    def _setup_data_interpolator(self) -> None:
```
jperson, line 242: 👍

jperson (approved):

PR #204, opened 2021-06-21
Initial websocket implementation through flask-socketio

tannermpeterson:


src/mantarray_desktop_app/main.py
```diff
@@ -258,8 +286,31 @@ def main(
...
-    server_thread.join()
+    logger.info("Starting Flask SocketIO")
+    _, host, _ = get_server_address_components()
+
+    data_queue_to_server = process_manager.queue_container().get_data_queue_to_server()
+
+    def data_sender() -> None:  # pragma: no cover  # Tanner (6/21/21): code coverage can't follow into start_background_task where this function is run
+        while True:
+            try:
+                item = data_queue_to_server.get(timeout=0.0001)
```
jperson, line 297: one design pattern that works well for this type of IPC is to have a 'tombstone' or sentinel message that tells the reader to shutdown instead of doing a busy poll on the queue.

I'm not sure if it would work in this case, but it might be worth considering

jperson (approved):

jperson (approved):

PR #224, opened 2021-07-30
New barcode format

tannermpeterson:


src/main/index.js
```diff
@@ -1,6 +1,6 @@
 /* globals INCLUDE_RESOURCES_PATH */
 import { app } from "electron";
-const autoUpdater = require("electron-updater"); // eslint-disable-line no-unused-vars
+const { autoUpdater } = require("electron-updater");
```
tannermpeterson, line 3: For some reason this fixes the "`autoUpdater.checkForUpdatesAndNotify` is not a function" issue

jperson: 👍

jperson (approved):

PR #262, opened 2021-10-22
Auto-upload feature

luciipak:


src/mantarray_desktop_app/file_writer.py
```diff
@@ -1006,6 +1032,162 @@ def _handle_performance_logging(self) -> None:
...
+                            "failed_upload": True,
+                            "customer_account_id": customer_dir,
+                            "thread": upload_thread,
+                            "auto_delete": False,
+                            "file_name": file_name,
+                        }
+                        self._upload_threads_container.append(thread_dict)
+                        # Lucy (10/22/2021): figure out how to handle if delete local files had been selected on original customer settings and how to handle the zip file
+
+    def _check_upload_statuses(self) -> None:
```
tannermpeterson, line 1151: This method is a great one for using the early exit pattern I mentioned in the other comment. `return`ing after line 1157 and `continue`ing after line 1165 would save two indents from the inner most branch of the loop. But of course it comes down to preference, both are fine

tests/file_writer/test_file_writer.py
```diff
@@ -117,11 +112,289 @@ def test_FileWriterProcess_soft_stop_not_allowed_if_incoming_data_still_in_queue
...
+    GENERIC_UPDATE_CUSTOMER_SETTINGS["config_settings"]["auto_delete_local_files"] = False
+    this_command = copy.deepcopy(GENERIC_UPDATE_CUSTOMER_SETTINGS)
+    put_object_into_queue_and_raise_error_if_eventually_still_empty(this_command, from_main_queue)
+
+    mocked_check_status = mocker.patch.object(file_writer_process, "_check_upload_statuses", auto_spec=True)
+    mocker.patch.object(file_uploader, "ErrorCatchingThread", autospec=True)
+
+    invoke_process_run_and_check_errors(file_writer_process)
+    # pylint: disable=protected-access
+    assert (
```
tannermpeterson, line 215: If you wanted to, you could make a getter method for `_upload_threads_container` so you don't have to disable pylint warnings. When I do that, I typically say in the docstring that the method is specifically for use in unit testing. Not sure which way is better TBH

tests/test_file_uploader.py
```diff
@@ -0,0 +1,323 @@
...
+    mocked_create_zip_file = mocker.patch.object(file_uploader, "create_zip_file", autospec=True)
+    mocker.patch.object(file_uploader, "get_file_md5", autospec=True)
+    mocker.patch.object(file_uploader, "get_access_token", autospec=True)
+    mocker.patch.object(file_uploader, "get_upload_details", autospec=True)
+    mocker.patch.object(file_uploader, "upload_file_to_s3", autospec=True)
+    mocker.patch.object(file_uploader, "get_sdk_status", autospec=True, return_value="analysis complete")
+
+    test_file_name = "zip_file"
+    test_file_path = "/test"
+    test_zip_dir = "/test/zipped_recordings"
```
tannermpeterson, line 138: This seems to work fine here, but sometimes you can run into issues when testing on different OS's when adding the separator character to a string directly since it's `/` on unix based OS's and `\` on windows. You can get around this issue by using `os.path.join("test", "zipped_recordings") which will insert the correct char depending on the OS

tannermpeterson (approved):

tests/test_file_uploader.py
luciipak: Yes! i actually had that error and had to swap a lot of them.. idk why this one didn't error

PR #266, opened 2021-11-08
upload status to main queue

luciipak:


tests/file_writer/test_recording.py
```diff
@@ -1724,3 +1660,96 @@ def test_FileWriterProcess__deletes_recorded_stim_data_after_stop_time(
...
+    assert (
+        file_writer_process._upload_threads_container[0]["auto_delete"] is False
+    )  # pylint: disable=protected-access
+    assert (
+        file_writer_process._upload_threads_container[0]["file_name"] == "MA200440001__2020_02_09_190935"
+    )  # pylint: disable=protected-access
+
+    invoke_process_run_and_check_errors(file_writer_process, num_iterations=8)
+
+    assert to_main_queue[-1]["communication_type"] == "update_upload_status"
```
tannermpeterson, line 1707: Indexing into the queue like this only works because a `TestingQueue` is used here which is just a modified deque, so this test would fail if someone changed it to a real queue. There's a helper function called `drain_queue` which will give you a list of all the items in the queue which is useful when you just want to assert something on the last item to be put into a queue

tests/fixtures_file_writer.py
```diff
@@ -347,3 +350,89 @@ def fixture_running_four_board_file_writer_process(runnable_four_board_file_writ
 
     fw_process.stop()
     fw_process.join()
+
+
+def file_writer_process_with_closed_h5_files_for_upload(
```
tannermpeterson, line 355: This name reads like a fixture but it is just a function that uses a `four_board_file_writer_process`. Could probably change it to something like `create_closed_h5_files_for_upload`. Something else to watch out for is that you actually run the process in this function. There is a `invoke_process_run_and_check_errors(file_writer_process, num_iterations=8)` in a test that uses this function and it's unclear at a glance what the number 8 does. If those iterations were done here then there would be no need for the user of this function to figure out how many loops to run file_writer. Also be sure to use the minimum number of iterations necessary (typically one iteration per item in the queue, can be more if that item causes something to happen on the next iteration though), and if applicable drain any queues to make assertions on them easier in tests

tests/file_writer/test_recording.py
```diff
@@ -1724,3 +1660,96 @@ def test_FileWriterProcess__deletes_recorded_stim_data_after_stop_time(
...
+    update_customer_settings_command = copy.deepcopy(GENERIC_UPDATE_CUSTOMER_SETTINGS)
+    update_customer_settings_command["config_settings"]["auto_delete_local_files"] = False
+    update_customer_settings_command["config_settings"]["auto_upload_on_completion"] = True
+    file_writer_process_ready_for_upload = file_writer_process_with_closed_h5_files_for_upload(
+        four_board_file_writer_process=four_board_file_writer_process,
+        update_customer_settings_command=update_customer_settings_command,
+    )
+    file_writer_process = file_writer_process_ready_for_upload["fw_process"]
+    to_main_queue = file_writer_process_ready_for_upload["to_main_queue"]
+    mocker.patch.object(file_uploader.ErrorCatchingThread, "errors", autospe=True, return_value=True)
```
tannermpeterson, line 1677: typo `autospec=True`
Should also mock the `start` function and anything else that will help to avoid sporadic failures

tests/file_writer/test_file_writer.py
```diff
@@ -289,46 +279,14 @@ def test_FileWriterProcess__prevent_any_uploads_with_no_stored_customer_settings
...
-    this_command = copy.deepcopy(GENERIC_UPDATE_CUSTOMER_SETTINGS)
-    put_object_into_queue_and_raise_error_if_eventually_still_empty(this_command, from_main_queue)
-    invoke_process_run_and_check_errors(file_writer_process)
-
-    assert (len(spied_delete_files.call_args_list) > 0) is auto_delete
-    assert len(file_writer_process._upload_threads_container) == 0  # pylint: disable=protected-access
-
-
 def test_FileWriterProcess__setup_before_loop__calls_super(four_board_file_writer_process, mocker):
     # pylint: disable=protected-access
```
tannermpeterson, line 291: Be careful not to dd pylint disables on their own line, that can have unintended behavior

tests/file_writer/test_recording.py
```diff
@@ -719,6 +652,9 @@ def test_FileWriterProcess__closes_the_files_and_adds_crc32_checksum_and_sends_c
     to_main_queue.get(
         timeout=QUEUE_CHECK_TIMEOUT_SECONDS
     )  # pop off the initial receipt of stop command message
+    to_main_queue.get(
+        timeout=QUEUE_CHECK_TIMEOUT_SECONDS
+    )  # pop off the initial receipt of stop command message
```
tannermpeterson, line 657: Update comment

tests/file_writer/test_recording.py
```diff
@@ -1724,3 +1660,96 @@ def test_FileWriterProcess__deletes_recorded_stim_data_after_stop_time(
...
+        file_writer_process._customer_settings == GENERIC_UPDATE_CUSTOMER_SETTINGS["config_settings"]
+    )  # pylint: disable=protected-access
+    assert (
+        file_writer_process._is_finalizing_files_after_recording() is False
+    )  # pylint: disable=protected-access
+    assert len(file_writer_process._open_files[0].keys()) == 0  # pylint: disable=protected-access
+    assert len(file_writer_process._upload_threads_container) == 1  # pylint: disable=protected-access
+    assert (
+        file_writer_process._upload_threads_container[0]["failed_upload"] is False
+    )  # pylint: disable=protected-access
```
tannermpeterson, line 1694: It looks like a lot of assertions are being made on this object, I would just make a getter for it in file_writer to avoid having the messy `# pylint: disable=protected-access` everywhere. One other thing is pylint disables need to actually be directly on the line where the thing that they're disabling is happening. These aren't actually doing anything here, I believe it's not causing any pylint errors though because of the protected-access disable on line 1683 which is on its own line

tannermpeterson (changes requested):
Looks good but small things to fix in testing. Also noticed the Docstring at the top of file_writer.py has never been updated, if you could change that to something more appropriate that would be great

tannermpeterson (approved):

PR #269, opened 2021-11-16
Upload file widget

luciipak:


tests/main/test_launch.py
```diff
@@ -432,7 +432,7 @@ def test_main__generates_log_file_uuid_if_none_passed_in_cmd_line_args(
     with tempfile.TemporaryDirectory() as tmp:
         test_dict = {
             "stored_customer_ids": {
-                "73f52be0-368c-42d8-a1fd-660d49ba5604": "filler_password",
+                "73f52be0-368c-42d8-a1fd-660d49ba5604": "Filler_password123",
```
tannermpeterson, line 435: Would probably be good to make these constants in the test file eventually

luciipak: I"ll do it when I add in the new screenshots after these run

tannermpeterson (approved):

PR #281, opened 2021-12-14
add usernames to /update_settings route and update s3 prefix with customer ID/username

luciipak:


package.json
```diff
@@ -30,7 +30,7 @@
     "org-imports": "organize-imports-cli src/main/index.dev.js"
   },
   "engines": {
-    "node": ">=14.15.1"
+    "node": "^14.15.1"
```
jperson, line 33: does the '^' pin it to the version?

jperson (approved):

package.json
luciipak: @jperson requires that major version

PR #292, opened 2021-12-16
user account refactor

luciipak:


src/mantarray_desktop_app/constants.py
```diff
@@ -70,9 +70,9 @@
 VALID_CONFIG_SETTINGS = frozenset(
     [
         "customer_account_uuid",
-        "user_account_uuid",
+        "user_account_id",
         "customer_pass_key",
-        "customer_username",
+        "user_account_id",
```
tannermpeterson, line 75: Can probably remove this second line since `user_account_id` is already in the set

tannermpeterson (approved):

src/mantarray_desktop_app/constants.py
luciipak: oh yes. my bad

PR #316, opened 2022-01-11
Upload log files

luciipak:


src/main/utils.js
```diff
@@ -49,8 +50,7 @@ const create_store = function ({
 
 const get_flask_logs_full_path = function (electron_store) {
   const electron_store_dir = path.dirname(electron_store.path);
-  const flask_logs_subfolder = "logs_flask";
-  return path.join(electron_store_dir, flask_logs_subfolder);
+  return path.join(electron_store_dir, "logs_flask", log_uuid);
```
luciipak, line 53: this is not be very user friendly for when they need to access the logs, they'll have to really pay attention to what the model says in order to get the correct uuid for that sessions logs. Think this be changed to `mantarray_logs__Y_M_D_Time` of the main process log instead of a uuid?

jperson: Yeah that's a good point, if we needed them to send us logs it would be tough to explain which ones to send. I like the `mantarray_logs_YYMMDD_Time` format

src/mantarray_desktop_app/log_formatter.py
```diff
@@ -13,13 +14,14 @@ class SensitiveFormatter(logging.Formatter):
...
 
     @staticmethod
-    def _filter(log_msg: str) -> str:
-        return re.sub(
-            r"(.*set_mantarray_nickname\?nickname=)(.*)( HTTP.*)",
-            lambda match_obj: match_obj[1] + get_redacted_string(len(match_obj[2])) + match_obj[3],
-            log_msg,
-        )
+    def _filter(log_msg: str) -> Any:
+        if "/update_settings" not in log_msg:
```
tannermpeterson, line 18: Why did this need to be added? I wouldn't think nicknames would ever show up in /update_settings

luciipak: @tannermpeterson I added that line so that the request with customer creds/password doesn't get logged. Doesn't have to do with the nickname.

tannermpeterson (approved):

PR #387, opened 2022-05-11
Mag algo addition + default layout redesign

luciipak:


src/mantarray_desktop_app/file_writer.py
```diff
@@ -196,8 +197,36 @@ def _drain_board_queues(
     return board_dict
 
 
+class FileDirMixIn(metaclass=abc.ABCMeta):
```
tannermpeterson, line 200: Since DataAnalyzer isn't using this MixIn at the moment, we should probably get rid of it and just keep all this functionality in FileWriter itself for now

src/mantarray_desktop_app/data_analyzer.py
```diff
@@ -550,6 +598,50 @@ def _process_stim_packet(self, stim_packet: Dict[Any, Any]) -> None:
...
+            "output_dir": self._mag_analysis_output_dir,
+        }
+
+    def _check_mag_analysis_statuses(self) -> None:
+        if thread_dict := self._mag_analysis_active_thread:
+            thread = thread_dict.get("thread")
+            recordings = thread_dict.get("recordings")
+            output_dir = thread_dict.get("output_dir")
+
+            if thread.is_alive():
```
tannermpeterson, line 623: Calling `join()` will block until the thread completes, so we want to change this to `if not thread.is_alive():` so that we know the thread is completed before joining it

src/mantarray_desktop_app/server.py
```diff
@@ -344,6 +345,59 @@ def boot_up() -> Response:
...
+
+    recording_paths = [
+        os.path.join(recording_rootdir, dirname) for dirname in request.get_json()["selected_recordings"]
+    ]
+
+    queue_command_to_main(
+        {
+            "communication_type": "mag_finding_analysis",
+            "command": "start_mag_analysis",
+            "content": {"recordings": recording_paths},
```
tannermpeterson, line 394: For these commands you don't need to wrap everything other than the comm type and command underneath a "content" key. Most of the commands coming from the server just look like
```
{ 
    "communication_type": "mag_finding_analysis",
    "command": "start_mag_analysis",
    "recordings": recording_paths
}
```

src/renderer/layouts/default.vue
```diff
@@ -214,13 +278,28 @@ export default {
...
-      this.confirmation_request = false;
+      this.$store.commit("settings/set_confirmation_request", false);
     },
     save_customer_id: function () {
       const customer_id = this.user_accounts[this.active_user_index].customer_id;
       ipcRenderer.send("save_customer_id", customer_id);
       this.$store.commit("settings/set_stored_customer_id", customer_id);
     },
+    handle_tab_visibility: function (tab) {
+      if (tab === 0) {
```
tannermpeterson, line 289: I think 
```
this.data_acquisition_visibility = tab === 0;
this.data_analysis_visibility = tab === 1;
this.stim_studio_visibility = tab === 2;
```
might work if you want to make this really compact

tests/data_analyzer/test_analysis.py
```diff
@@ -309,7 +309,7 @@ def test_check_for_new_twitches__returns_latest_twitch_index_and_populated_metri
 def test_DataAnalyzerProcess__sends_beta_1_metrics_of_all_wells_to_main_when_ready(
     four_board_analyzer_process, mantarray_mc_simulator, mocker
 ):
-    da_process, board_queues, from_main_queue, _, _ = four_board_analyzer_process
+    da_process, board_queues, from_main_queue, _, _, _ = four_board_analyzer_process
```
tannermpeterson, line 312: Could use `da_process, board_queues, from_main_queue, *_`, but we really need to change the `four_board_analyzer_process` fixture to be a dict instead of tuple

tests/data_analyzer/test_beta_2_mode.py
```diff
@@ -340,3 +342,122 @@ def test_DataAnalyzerProcess__formats_and_passes_incoming_stim_packet_through_to
...
+
+    test_mag_analysis_command = copy.deepcopy(TEST_START_MAG_ANALYSIS_COMMAND)
+
+    put_object_into_queue_and_raise_error_if_eventually_still_empty(
+        test_mag_analysis_command, from_main_queue
+    )
+    invoke_process_run_and_check_errors(da_process, 1)
+    spied_start_analysis.assert_called_once_with(TEST_START_MAG_ANALYSIS_COMMAND["content"])
+
+    drain_queue(from_main_queue)
```
tannermpeterson, line 365: `invoke_process_run_and_check_errors(da_process, 1)` will grab the command out of `from_main_queue`, so you don't need to drain it here

tests/data_analyzer/test_beta_2_mode.py
```diff
@@ -340,3 +342,122 @@ def test_DataAnalyzerProcess__formats_and_passes_incoming_stim_packet_through_to
...
+    mocker.patch.object(PlateRecording, "write_time_force_csv", side_effect=Exception("exception raised"))
+
+    test_mag_analysis_command = copy.deepcopy(TEST_START_MAG_ANALYSIS_COMMAND)
+    put_object_into_queue_and_raise_error_if_eventually_still_empty(
+        test_mag_analysis_command, from_main_queue
+    )
+    invoke_process_run_and_check_errors(da_process, 7)
+    assert len(spied_check_analysis.call_args_list) == 7
+
+    msg = to_main_queue.get()
```
tannermpeterson, line 411: use `get_nowait()` or `get(timeout=QUEUE_CHECK_TIMEOUT_SECONDS)` since `get()` will block if there are no items in the queue

tests/data_analyzer/test_beta_2_mode.py
```diff
@@ -340,3 +342,122 @@ def test_DataAnalyzerProcess__formats_and_passes_incoming_stim_packet_through_to
...
+    mocked_pr = mocker.patch.object(PlateRecording, "from_directory", autospec=True)
+    mocked_pr.return_value = iter([PlateRecording])
+    mocker.patch.object(next(mocked_pr.return_value), "write_time_force_csv", autospec=True)
+
+    test_mag_analysis_command = copy.deepcopy(TEST_START_MAG_ANALYSIS_COMMAND)
+    put_object_into_queue_and_raise_error_if_eventually_still_empty(
+        test_mag_analysis_command, from_main_queue
+    )
+    invoke_process_run_and_check_errors(da_process, 7)
+    msg = to_main_queue.get()
```
tannermpeterson, line 448: ditto

tests/fixtures.py
```diff
@@ -50,10 +50,18 @@ def generate_board_and_error_queues(num_boards: int = 4, queue_type=MPQueue):
...
 
-def get_generic_base64_args(recording_directory: Optional[str] = None) -> str:
+def get_generic_base64_args(
+    recording_directory: Optional[str] = None, analysis_output_dir: Optional[str] = None
+) -> str:
     if not recording_directory:
         recording_directory = get_current_file_abs_directory()
-    test_dict = {"log_file_id": "any", "recording_directory": recording_directory}
+    if not analysis_output_dir:
+        analysis_output_dir = get_current_file_abs_directory()
```
tannermpeterson, line 59: We probably want to make these file dirs different just in case a test tries to use both of them. Could just use
```
recording_directory = os.path.join(get_current_file_abs_directory(), "recordings")
analysis_output_dir = os.path.join(get_current_file_abs_directory(), "time_force_data")
```

tests/fixtures_data_analyzer.py
```diff
@@ -9,6 +11,15 @@
 from .fixtures import QUEUE_CHECK_TIMEOUT_SECONDS
 from .helpers import put_object_into_queue_and_raise_error_if_eventually_still_empty
 
+TEST_RECORDING_ROOT_PATH = os.path.join(os.path.abspath("tests"), "test_recording_h5")
+TEST_REC_DIR_PATH = os.path.join(TEST_RECORDING_ROOT_PATH, "ML2021172153__2022_01_21_023323")
```
tannermpeterson, line 15: Could probably use `tempfile` to create a `TemporaryFile` wherever a real file is needed instead of committing an empty file to the repo just for rest purposes

tannermpeterson (changes requested):

src/mantarray_desktop_app/data_analyzer.py
luciipak: yeah this was a major typo. I had switched it to something ele and forgot that part

src/mantarray_desktop_app/server.py
luciipak: cool, i'll change it. Initially did that in case we ever needed more info bc I was passing the output directory here, but now what you're saying makes more sense

tannermpeterson (approved):

src/renderer/layouts/default.vue
tannermpeterson: Oops, it should actually be this
```
this.data_acquisition_visibility = tab === 0 && !this.data_acquisition_visibility;
this.stim_studio_visibility = tab === 1 && !this.stim_studio_visibility;
this.data_analysis_visibility = tab === 2 && !this.data_analysis_visibility;
```
I'll fix it in my next PR though

PR #400, opened 2022-06-24
added h5 corruption check

NikitaPoly:


tests/file_writer/test_recording.py
```diff
@@ -790,6 +788,60 @@ def test_FileWriterProcess__closes_the_files_and_sends_communication_to_main_whe
...
 
+def test_FileWriterProcess__sends_message_to_main_if_a_corrupt_file_is_found(
+    four_board_file_writer_process, mocker
+):
+    file_dir = four_board_file_writer_process["file_dir"]
+    to_main_queue = four_board_file_writer_process["to_main_queue"]
+
+    test_num_data_points = 10
+
+    def file_se(file_name, open_type, *args, **kwargs):
```
tannermpeterson, line 799: Something I forgot to mention is that my general rule of thumb for comments is if it's not immediately clear why something is being done, or what's even being done, it deserves a comment. Personally I find this happens more for me in test code I write, especially when mocking, but it can certainly happen in source code as well. This is a pretty funky side effect so it definitely deserves a comment, and something that might be good is "mocking here so that the call to this function when creating the file is fine but the second call when checking for corruption raises an error"

tannermpeterson:
merged branch into v1-mappings

PR #414, opened 2022-08-10
Recording snapshot feature

luciipak:


src/mantarray_desktop_app/sub_processes/file_writer.py
```diff
@@ -459,9 +459,11 @@ def _process_next_command_from_main(self) -> None:
             self._process_update_name_command(communication)
             to_main.put_nowait(
                 {
-                    "communication_type": "command_receipt",
+                    "communication_type": "mag_finding_analysis"
```
tannermpeterson, line 462: Instead of changing the comm type, you could just add the `"snapshot_enabled"` value in the command response sent back to main, but I think either is fine

tannermpeterson (approved):

PR #435, opened 2022-11-03
Updated upload type and login response

luciipak:


tannermpeterson (approved):

tannermpeterson:
We still need to bring in the new FE components changes, but that can happen in a separate PR

PR #441, opened 2022-11-15
remove spaces from start and end of new file name

NikitaPoly:
new files have leading and ending spaces removed.

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -48,6 +48,7 @@ streamz = "==0.6.2"
 wakepy = "==0.5.0"
 werkzeug = "==2.0.3"  # Tanner (4/7/22): pinned for flask
 xem-wrapper = "==0.3.0"
+mantarray-desktop-app = {editable = true, path = "."}
 
 [requires]
 python_version = "3.8"
--- a/Pipfile
+++ b/Pipfile
@@ -48,7 +48,6 @@ streamz = "==0.6.2"
 wakepy = "==0.5.0"
 werkzeug = "==2.0.3"  # Tanner (4/7/22): pinned for flask
 xem-wrapper = "==0.3.0"
-mantarray-desktop-app = {editable = true, path = "."}
 
 [requires]
 python_version = "3.8"
--- a/src/mantarray_desktop_app/sub_processes/file_writer.py
+++ b/src/mantarray_desktop_app/sub_processes/file_writer.py
@@ -1257,7 +1257,6 @@ class FileWriterProcess(InfiniteProcess):
         """Rename recording directory and h5 files to kick off auto upload."""
         # only perform if new name is different from the original default name
         if self._current_recording_dir == comm["default_name"] != comm["new_name"]:
-            comm["new_name"] = comm["new_name"].strip()
             new_recording_path = os.path.join(self._file_directory, comm["new_name"])
             if os.path.exists(new_recording_path):
                 # remove current recording if it already exists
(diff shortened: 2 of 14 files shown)


Pipfile
```diff
@@ -48,6 +48,7 @@ streamz = "==0.6.2"
 wakepy = "==0.5.0"
 werkzeug = "==2.0.3"  # Tanner (4/7/22): pinned for flask
 xem-wrapper = "==0.3.0"
+mantarray-desktop-app = {editable = true, path = "."}
```
tannermpeterson, line 51: I don't think we want this here. If you added this for local development, the correct way to do it is to just run `pipenv run pip install -e .` similar to what we do for development in other repos

src/mantarray_desktop_app/sub_processes/file_writer.py
```diff
@@ -1257,6 +1257,7 @@ def _process_update_name_command(self, comm: Dict[str, str]) -> None:
         """Rename recording directory and h5 files to kick off auto upload."""
         # only perform if new name is different from the original default name
         if self._current_recording_dir == comm["default_name"] != comm["new_name"]:
+            comm["new_name"] = comm["new_name"].strip()
```
tannermpeterson, line 1260: We might want to strip this in the route that receives this new name. Also we always want to unit test changes we make. I know we aren't doing that in the cloud right now, but pushing updates for the desktop app is much more work than in the cloud so it's much more important to make sure everything is adequately tested before we ship

tannermpeterson (changes requested):

Pipfile
NikitaPoly: I ran the commands from confluence to update my repo. This was added from the script.

tests/server/test_calling_routes.py
```diff
@@ -485,6 +485,25 @@ def test_update_recording_name__returns_ok_if_recording_name_doesnt_exist(
     assert response.status_code == 200
 
 
+@pytest.mark.parametrize(
+    "test_new_file_name",
+    ["NormalName", "  TestName", "TestName  ", "  TestName  "],
+)
+def test_update_recording_name__trims_spaces_from_start_and_end_beta_2_mode(
```
tannermpeterson, line 492: nitpick, you don't need to have beta_2_mode mentioned in the test name now that there is no beta 1 mode version of this test

tests/server/test_calling_routes.py
```diff
@@ -485,6 +485,25 @@ def test_update_recording_name__returns_ok_if_recording_name_doesnt_exist(
...
+    test_client, _, shared_values_dict = client_and_server_manager_and_shared_values
+    shared_values_dict["beta_2_mode"] = True
+    shared_values_dict["config_settings"]["recording_directory"] = "/test/recording/directory"
+
+    mocker.patch.object(os.path, "exists", return_value=False)
+
+    response = test_client.post(
+        f"/update_recording_name?new_name={test_new_file_name}&default_name=old_name&snapshot_enabled=true"
+    )
+    assert response.status_code == 200
```
tannermpeterson, line 504: The purpose of this test is to make sure that the new file name gets trimmed, but the only assertion made is this one. We need to make sure that the message this route sends to `main` contains the trimmed file name. You should be able to find some other tests that do this in `tests/server/test_routes_populate_queues`, I also think it makes sense to move this test into that file

tannermpeterson (approved):

PR #448, opened 2022-12-04
Stim protocol updates no chunks

tannermpeterson:


src/mantarray_desktop_app/sub_processes/file_writer.py
```diff
@@ -673,7 +676,8 @@ def _process_start_recording_command(self, communication: Dict[str, Any]) -> Non
         if self._beta_2_mode:
             stim_data_buffers = self._stim_data_buffers[board_idx]
             for well_idx, well_buffers in stim_data_buffers.items():
-                self._handle_recording_of_stim_statuses(well_idx, np.array(well_buffers))
+                if well_buffers[0]:
```
luciipak, line 679: won't this cause an index error if it's empty?

luciipak (approved):

PR #464, opened 2022-12-20
end_time in mag finding thread to default to none

luciipak:


src/mantarray_desktop_app/workers/magnet_finder.py
```diff
@@ -18,7 +17,7 @@ def run_magnet_finding_alg(
     result_dict: Dict[str, Any],
     recordings: List[str],
     output_dir: Optional[str] = None,
-    end_time: Union[float, int] = np.inf,
+    end_time: Union[float, int] = None,
```
tannermpeterson, line 20: I think you're going to need to make this `Optional[Union[float, int]]` or mypy will complain

tannermpeterson (approved):

src/mantarray_desktop_app/sub_processes/data_analyzer.py
```diff
@@ -652,7 +652,9 @@ def _check_mag_analysis_statuses(self) -> None:
     def _start_recording_snapshot_analysis(self, recording_path: str) -> None:
         # TODO (9/16/22): this should be run in a thread so that this process is still responsive to main
         # TODO (11/16/22): add error handling. If the analysis fails here, then snapshot_dfs[0] will raise a KeyError
-        snapshot_dfs = run_magnet_finding_alg({}, [recording_path], end_time=RECORDING_SNAPSHOT_DUR_SECS)
+        snapshot_dfs = run_magnet_finding_alg(
```
tannermpeterson, line 655: A test is failing because of the None here, but I don't think you need to have that in this call since it's the default value

luciipak: oh yeah, meant to remove that. i added it when I removed the default since it's optional, but then errored about positional argument placements

tannermpeterson (approved):

PR #496, opened 2023-02-23
redact username and password from logs

NikitaPoly:


src/mantarray_desktop_app/utils/generic.py
```diff
@@ -148,17 +148,39 @@ def convert_request_args_to_config_dict(request_args: Dict[str, Any]) -> Dict[st
     return config_dict
 
 
-def redact_sensitive_info_from_path(file_path: Optional[str]) -> Optional[str]:
-    """Scrubs username from file path to protect sensitive info."""
-    if file_path is None:
+def redact_sensitive_info(
```
tannermpeterson, line 151: This isn't quite what we want, this method shouldn't be changed at all. We want a signle  method that you can pass an item into that will handle all possible redaction cases, whether it be a dict with a sensitive field, or a string with some sensitive info, let's just call it `redact_sensitive_info`. This method will call `redact_sensitive_info_from_path` when it encounters a file/folder path. We also need to move the loop further down back where it originally was in process monitor. Once the loop is moved back, you'll just be calling `redact_sensitive_info` on each item

tests/process_monitor/test_comm_from_server.py
```diff
@@ -664,15 +664,40 @@ def test_MantarrayProcessesMonitor__check_and_handle_server_to_main_queue__handl
...
..., 0, 0, 0, 5, 0, 5, 0, 5, 0, 0, 0, 5, 0, 5, 0, 5, 0, 0, 0, 5, 0, 5, 0, 5, 0, 0, 128, 4, 128, 4, 128, 4, 0, 0, 128, 4, 128, 4, 128, 4, 0, 0, 128, 4, 128, 4, 128, 4, 0, 0, 0, 8, 0, 8, 0, 8, 0, 0, 0, 8, 0, 8, 0, 8, 0, 0, 0, 8, 0, 8, 0, 8, 0, 0, 128, 7, 128, 7, 128, 7, 0, 0, 128, 7, 128, 7, 128, 7, 0, 0, 128, 7, 128, 7, 128, 7, 0, 0, 0, 7, 0, 7, 0, 7, 0, 0, 0, 7, 0, 7, 0, 7, 0, 0, 0, 7, 0, 7, 0, 7, 0, 0, 128, 6, 128, 6, 128, 6, 0, 0, 128, 6, 128, 6, 128, 6, 0, 0, 128, 6, 128, 6, 128, 6, 0, 0, 0, 10, 0, 10, 0, 10, 0, 0, 0, 10, 0, 10, 0, 10, 0, 0, 0, 10, 0, 10, 0, 10, 0, 0, 128, 9, 128, 9, 128, 9, 0, 0, 128, 9, 128, 9, 128, 9, 0, 0, 128, 9, 128, 9, 128, 9, 0, 0, 0, 9, 0, 9, 0, 9, 0, 0, 0, 9, 0, 9, 0, 9, 0, 0, 0, 9, 0, 9, 0, 9, 0, 0, 128, 8, 128, 8, 128, 8, 0, 0, 128, 8, 128, 8, 128, 8, 0, 0, 128, 8, 128, 8, 128, 8, 0, 0, 0, 12, 0, 12, 0, 12, 0, 0, 0, 12, 0, 12, 0, 12, 0, 0, 0, 12, 0, 12, 0, 12, 0, 0, 128, 11, 128, 11, 128, 11, 0, 0, 128, 11, 128, 11, 128, 11, 0, 0, 128, 11, 128, 11, 128, 11, 0, 0, 0, 11, 0, 11, 0, 11, 0, 0, 0, 11, 0, 11, 0, 11, 0, 0, 0, 11, 0, 11, 0, 11, 0, 0, 128, 10, 128, 10, 128, 10, 0, 0, 128, 10, 128, 10, 128, 10, 0, 0, 128, 10, 128, 10, 128, 10, 34, 226, 12, 164]",
```
tannermpeterson, line 686: We shouldn't be adding strings this large into tests. Also, this wouldn't need any redaction, we want to add messages here that _will_ need redaction, and maybe only a single one that doesn't

tests/process_monitor/test_process_monitor.py
```diff
@@ -518,13 +518,44 @@ def test_MantarrayProcessesMonitor__logs_errors_from_data_analyzer(
...
..., 0, 0, 0, 5, 0, 5, 0, 5, 0, 0, 0, 5, 0, 5, 0, 5, 0, 0, 0, 5, 0, 5, 0, 5, 0, 0, 128, 4, 128, 4, 128, 4, 0, 0, 128, 4, 128, 4, 128, 4, 0, 0, 128, 4, 128, 4, 128, 4, 0, 0, 0, 8, 0, 8, 0, 8, 0, 0, 0, 8, 0, 8, 0, 8, 0, 0, 0, 8, 0, 8, 0, 8, 0, 0, 128, 7, 128, 7, 128, 7, 0, 0, 128, 7, 128, 7, 128, 7, 0, 0, 128, 7, 128, 7, 128, 7, 0, 0, 0, 7, 0, 7, 0, 7, 0, 0, 0, 7, 0, 7, 0, 7, 0, 0, 0, 7, 0, 7, 0, 7, 0, 0, 128, 6, 128, 6, 128, 6, 0, 0, 128, 6, 128, 6, 128, 6, 0, 0, 128, 6, 128, 6, 128, 6, 0, 0, 0, 10, 0, 10, 0, 10, 0, 0, 0, 10, 0, 10, 0, 10, 0, 0, 0, 10, 0, 10, 0, 10, 0, 0, 128, 9, 128, 9, 128, 9, 0, 0, 128, 9, 128, 9, 128, 9, 0, 0, 128, 9, 128, 9, 128, 9, 0, 0, 0, 9, 0, 9, 0, 9, 0, 0, 0, 9, 0, 9, 0, 9, 0, 0, 0, 9, 0, 9, 0, 9, 0, 0, 128, 8, 128, 8, 128, 8, 0, 0, 128, 8, 128, 8, 128, 8, 0, 0, 128, 8, 128, 8, 128, 8, 0, 0, 0, 12, 0, 12, 0, 12, 0, 0, 0, 12, 0, 12, 0, 12, 0, 0, 0, 12, 0, 12, 0, 12, 0, 0, 128, 11, 128, 11, 128, 11, 0, 0, 128, 11, 128, 11, 128, 11, 0, 0, 128, 11, 128, 11, 128, 11, 0, 0, 0, 11, 0, 11, 0, 11, 0, 0, 0, 11, 0, 11, 0, 11, 0, 0, 0, 11, 0, 11, 0, 11, 0, 0, 128, 10, 128, 10, 128, 10, 0, 0, 128, 10, 128, 10, 128, 10, 0, 0, 128, 10, 128, 10, 128, 10, 34, 226, 12, 164]",
```
tannermpeterson, line 546: ditto

tannermpeterson (changes requested):

PR #522, opened 2023-04-18
version bump for front end components to 1.4.7

NikitaPoly:


package.json
```diff
@@ -35,7 +35,7 @@
     "@types/node": "16.15.1"
   },
   "dependencies": {
-    "@curi-bio/mantarray-frontend-components": "1.4.6",
+    "@curi-bio/mantarray-frontend-components": "1.4.7",
```
tannermpeterson, line 38: You'll need to actually run `npm i` so that the lock file updates too

tannermpeterson (approved):

PR #537, opened 2023-05-04
New user login flow

luciipak:


jperson (approved):

NikitaPoly (approved):

src/mantarray_desktop_app/main_process/server.py
```diff
@@ -409,17 +405,41 @@ def update_settings() -> Response:
     """Update the customer/user settings.
 
     Can be invoked by: curl http://localhost:4567/update_settings?customer_id=<UUID>&user_name=<UUID>&recording_directory=<recording_dir>
-                       curl http://localhost:4567/update_settings?customer_id=<string>&user_password=<string>&user_name=<string>&auto_upload=<bool>&auto_delete=<bool>
+                       curl http://localhost:4567/update_settings?auto_upload=<bool>&auto_delete=<bool>&pulse3d_version=<str>
```
tannermpeterson, line 408: Don't worry about updating these anymore. I think we should just delete them now since no one is going to be hitting these routes from curl anymore

PR #551, opened 2023-05-25
Stim subprotocol looping

luciipak:


src/mantarray_desktop_app/main_process/server.py
```diff
@@ -484,80 +479,26 @@ def set_protocols() -> Response:
...
-                            "interphase_interval": lambda n: n >= 0,
-                        }
-                    )
-                for component_name, validator in subprotocol_component_validators.items():
-                    if not validator(component_value := subprotocol[component_name]):  # type: ignore
-                        component_name = component_name.replace("_", " ")
-                        return Response(
-                            status=f"400 Protocol {protocol_id}, Subprotocol {idx}, Invalid {component_name}: {component_value}"
+        try:
+            for idx, subprotocol in enumerate(protocol["subprotocols"]):
```
tannermpeterson, line 483: I think we should use recursion instead of looping in case we need to add more layers of nesting

NikitaPoly (dismissed):

NikitaPoly (dismissed):

tannermpeterson (dismissed):

tannermpeterson (approved):

PR #567, opened 2023-06-21
removed /system_status logging

NikitaPoly:
`/system_status` only gets logged when 200 status code is not present

[mantarray_log__2023_06_28_225724.txt](https://github.com/CuriBio/mantarray-desktop-app/files/11900521/mantarray_log__2023_06_28_225724.txt)


src/mantarray_desktop_app/main.py
```diff
@@ -127,6 +127,13 @@ def _log_system_info() -> None:
         logger.info(msg)
 
 
+class Filter_System_Status_Logs(logging.Filter):
```
tannermpeterson, line 130: We're going to remove this class most likely since this isn't the correct way to handle logging related to flask routes, but keep in mind this is the incorrect casing for python classes. Please check the definitions of other classes in this or other python repos for the correct casing.

src/mantarray_desktop_app/main.py
```diff
@@ -138,6 +145,9 @@ def main(command_line_args: List[str], object_access_for_testing: Optional[Dict[
         # Tanner (5/20/22): not sure if this is actually logging anything since logging isn't configured yet
         logger.info(command_line_args)
 
+        # add filtering of /system_status logs
+        logger.addFilter(Filter_System_Status_Logs())
```
tannermpeterson, line 149: A couple things here:
1. We already have a log formatter that we use. If we need to add any more filtering, it should be added to that one. I've mentioned this before, but it's very important that you read through and consider all of the existing code when adding new features to avoid adding something that is already present.
2. This solution does not filter out the log line that starts with `Response to HTTP Request in next log entry:`. You'll need to make sure we don't log that either for /system_status calls that return 200. Going of off the previous point, you will find that there is an existing function that can be modified to do this

luciipak:
Could you also attach a log file with these changes applied?

src/mantarray_desktop_app/main.py
```diff
@@ -127,6 +127,13 @@ def _log_system_info() -> None:
         logger.info(msg)
 
 
+class Filter_System_Status_Logs(logging.Filter):
+    def filter(self: Filter_System_Status_Logs, record: logging.LogRecord) -> bool:
+        # log system_status only if not 200
+        message = record.getMessage()
+        return not ("system_status" in message and " 200 " in message)
```
tannermpeterson, line 134: What happens if 200 is in this log line but it's not the return code?

tannermpeterson (changes requested):

CHANGELOG.rst
```diff
@@ -8,6 +8,10 @@ Fixed:
 ^^^^^^
 - Heatmap values to be converted from N to µN before sending to FE.
 
+Changed:
+^^^^^^^^
+- Limited logging of /system_status to only when status code is not 200
```
tannermpeterson, line 13: This is too technical for our users. It should be something higher level that just mentions that unnecessary log lines have been remove from log files

src/mantarray_desktop_app/main.py
tannermpeterson: Also remember to add tests. One of them should test this edge case

NikitaPoly:
> Could you also attach a log file with these changes applied?

Attached logs to the description of this PR @luciipak

src/mantarray_desktop_app/main.py
NikitaPoly: Initially I added spaces between 200 " 200 " to look for a 200 status code but I changed it to look in a specific location for the error code and check if it is 200 or not.

CHANGELOG.rst
NikitaPoly: changed to 

```
Removed:
^^^^^^^^
- Unnecessary logging from log files
```

src/mantarray_desktop_app/main.py
NikitaPoly: I implemented your comments but can't find the  `Response to HTTP Request in next log entry:`  `/system` log you mentioned. Is it in the filtered log file I posted?

CHANGELOG.rst
tannermpeterson: We don't use a `Removed` subsection, we have `Added`, `Changed`, and `Fixed`. If you look through the changelog I believe all removals go under `Changed`

src/mantarray_desktop_app/utils/log_formatter.py
```diff
@@ -15,7 +15,12 @@ class SensitiveFormatter(logging.Formatter):
 
     @staticmethod
     def _filter(log_msg: str) -> Any:
-        if "/update_settings" not in log_msg:
+        if "/system_status" in log_msg:
+            parts = log_msg.split()
```
tannermpeterson, line 19: This should use a regex like the filter for `/update_settings`

controller/tests/server/test_route_processing.py
```diff
@@ -1283,31 +1283,6 @@ def test_set_protocols__waits_for_stim_info_in_shared_values_dict_to_be_updated_
     assert mocked_sleep.call_args_list == [mocker.call(0.1), mocker.call(0.1)]
 
 
-def test_after_request__redacts_mantarray_nicknames_from_system_status_log_message(
```
tannermpeterson, line 1286: Why did we remove this test?

CHANGELOG.rst
```diff
@@ -8,6 +8,10 @@ Fixed:
 ^^^^^^
 - Heatmap values to be converted from N to µN before sending to FE.
 
+Changed:
+^^^^^^^^
+- Removed unnecessary logs lines from log files.
```
tannermpeterson, line 13: Should be "log lines" not "logs lines"

controller/tests/server/test_route_processing.py
```diff
@@ -1382,6 +1357,30 @@ def test_after_request__redacts_recording_folder_path_from_get_recordings_log_me
...
+
+def test_after_request__status_code_system_status_error_logged(
+    client_and_server_manager_and_shared_values, test_client, mocker
+):
+    *_, shared_values_dict = client_and_server_manager_and_shared_values
+    shared_values_dict["system_status"] = CALIBRATED_STATE
+    spied_server_logger = mocker.spy(server.logger, "info")
+    response = test_client.put("/system_status")
+
+    assert response.status_code != 200
```
tannermpeterson, line 1380: This test is identical to the one above however you are checking that the status code is not 200. You'll need to actually mock this route or configure the shared values so that this route does not return 200, otherwise this test will always fail

controller/tests/server/test_route_processing.py
NikitaPoly: This test checks that `systems status` logs are redacted if the status code is 200. If we are no longer logging 200 status code `system_status` logs, but instead the error , the logs look completely different and don't contain the message to redact? So I removed this test and replaced it with two tests, one to check logs on a 200 code and one to check logs on a none 200 status code.

  ```
if response.status_code == 200 and "system_status" not in rule.rule:
        # Tanner (1/19/21): using json.dumps instead of an f-string here allows us to perform better testing of our log messages by loading the json string to a python dict
        msg += json.dumps(response_json)
        logger.info(msg)
    elif response.status_code != 200:
        msg += response.status
        logger.info(msg)
    return response
```

Comment (NikitaPoly) [reply controller/tests/server/test_route_processing.py:1286]: @tannermpeterson am I understanding the test incorrectly?

Comment (tannermpeterson) [reply controller/tests/server/test_route_processing.py:1286]: I believe you understanding it correctly. I'm just asking a question to make sure that the test is no longer necessary

File: controller/src/mantarray_desktop_app/main_process/server.py [hunk 9]
```diff
@@ -1104,12 +1100,13 @@ def after_request(response: Response) -> Response:
                 response_json["root_recording_path"]
             )
     msg = "Response to HTTP Request in next log entry: "
-    if response.status_code == 200:
+    if response.status_code == 200 and "system_status" not in rule.rule:
```
tannermpeterson, line 1103: This check is identical to the one above. This is a good hint that the branches should either be combined or there is superfluous logic in one of them

controller/src/mantarray_desktop_app/main_process/server.py
```diff
@@ -1104,12 +1100,13 @@ def after_request(response: Response) -> Response:
...
         # Tanner (1/19/21): using json.dumps instead of an f-string here allows us to perform better testing of our log messages by loading the json string to a python dict
         msg += json.dumps(response_json)
-    else:
+        logger.info(msg)
+    elif response.status_code != 200:
         msg += response.status
-    logger.info(msg)
+        logger.info(msg)
```
tannermpeterson, line 1109: We now how 2 lines identical lines: `logger.info(msg)`. It would be better to just have it once and guard it with and `if` if necessary

controller/tests/server/test_route_processing.py
NikitaPoly: The second test makes a PUT call to the server instead of a GET call. This returns an error code that is not 200 and logs `/system_status`.  `spied_server_logger` is then used to check if any logs have been made. When testing `system_status` filtering,  a normal GET request is made, but this time `spied_server_logger` should not return any logs being made.

I can combine these two tests with parametrize.

What would be a some other ways to trigger an error other than changing the request method?

controller/src/mantarray_desktop_app/main_process/server.py
```diff
@@ -1104,11 +1100,13 @@ def after_request(response: Response) -> Response:
...
-    if response.status_code == 200:
+    if response.status_code == 200 and "system_status" not in rule.rule:
         # Tanner (1/19/21): using json.dumps instead of an f-string here allows us to perform better testing of our log messages by loading the json string to a python dict
         msg += json.dumps(response_json)
-    else:
+    elif response.status_code != 200:
         msg += response.status
+    else:
+        return response
     logger.info(msg)
```
tannermpeterson, line 1110: This will still log /system_status. What I'm getting at is that you only really need to do this here
```
if not ("system_status" in rule.rule and response.status_code == 200):
    logger.info(msg)
```
The `and "system_status" not in rule.rule:` is unnecessary

controller/tests/server/test_route_processing.py
```diff
@@ -1382,6 +1357,28 @@ def test_after_request__redacts_recording_folder_path_from_get_recordings_log_me
     }
 
 
+@pytest.mark.parametrize(
+    "mocked_request_method,expected_status_code,is_expecting_logs", [("GET", 200, False), ("PUT", 404, True)]
```
tannermpeterson, line 1361: This technically works, but /system_status doesn't allow `PUT` so like I mentioned before we should either use mocking or set up `shared_values_dict` to get /system_status to return something other than 200

tannermpeterson (dismissed):

tannermpeterson (approved):

PR #579, opened 2023-07-07
allow 300 barcodes

luciipak:


tannermpeterson (commented):
We should still try to get the pulse3d fix in

luciipak:
> We should still try to get the pulse3d fix in

that'll have to be done separately, just posted the PR for that. I'll add 0.33.10 into K8s

tannermpeterson:
The amplitude in live view will be incorrect until we bring pulse3d 0.33.11 into the MA Controller

luciipak:
> The amplitude in live view will be incorrect until we bring pulse3d 0.33.11 into the MA Controller

Oh duh, yeah forgot it's a dependency in this repo

tannermpeterson (approved):

PR #593, opened 2023-07-28
Limit failed login attempts

luciipak:


controller/src/mantarray_desktop_app/exceptions.py
```diff
@@ -290,8 +291,8 @@ class CloudAnalysisJobFailedError(Exception):
 class CloudAuthFailedError(Exception):
     """Base class for cloud auth related errors."""
 
-    def __init__(self, status_code: int):
-        super().__init__(f"Status Code: {status_code}")
+    def __init__(self, status_code: int, error_msg: Optional[str] = None):
+        super().__init__(f"Status code {status_code}: {error_msg}")
```
tannermpeterson, line 295: nitpick, shouldn't include that colon or error msg at the end if the error msg is None

tannermpeterson (approved):

controller/src/mantarray_desktop_app/exceptions.py
luciipak: @tannermpeterson fixed

tannermpeterson (approved):

PR #628, opened 2023-12-22
View changelog

luciipak:


jperson (approved):

tannermpeterson:
Do we need this to work for people who are not connected to the internet?

luciipak:
> Do we need this to work for people who are not connected to the internet?

As in show them the changelog offline? Currently will just open their default browser and if not connected, the browser will handle it

PR #656, opened 2024-11-08
Allow testing p3d versions

tannermpeterson:


electron/main/index.js
```diff
@@ -153,11 +153,9 @@ ipcMain.once("pulse3d_versions_request", (event) => {
   axios
     .get(`https://${CLOUD_PULSE3D_ENDPOINT}/versions`)
     .then((response) => {
-      const versions = response.data
-        .filter(({ state }) => state === "external")
-        .map(({ version }) => version);
```
roberto-araiza-curibio, line 158: qq: we don't do this `.map` here (1/2)

ui/components/settings/SettingsForm.vue
```diff
@@ -242,16 +243,37 @@ export default {
...
+        mapped_versions[info.version] = info;
+      });
+      const external_versions = semver_sort.desc(
+        this.pulse3d_version_info.filter(({ state }) => state === "external").map(({ version }) => version)
+      );
+      const testing_versions = semver_sort.desc(
+        this.pulse3d_version_info.filter(({ state }) => state === "testing").map(({ version }) => version)
+      );
+      const deprecated_versions = semver_sort.desc(
+        this.pulse3d_version_info.filter(({ state }) => state === "deprecated").map(({ version }) => version)
```
roberto-araiza-curibio, line 266: because we're doing these `.map`s here? (2/2)

roberto-araiza-curibio (approved):
LGTM.  One minor question (in two parts) below.

ui/components/settings/SettingsForm.vue
tannermpeterson: Yes, we need to store all the information that comes back from that response and filter later

